PR #16, opened 2023-03-07
Change users

### LINKED ISSUE #11: Create endpoints for creating, editing, and deleting users
### Goal: to create endpoints for creating, editing, and deleting users
**Branch name:** `change-users`

### Task at Hand
1. View the API documentation on our Notion page
2. Open `user/views.ts` (NOT `users/views.ts`) and define the following endpoints: `/user/create`, `/user/delete/`, and `/user/update`. The endpoint responses should follow the status codes written in the API documentation.
3. Open `user/controllers.ts` and write functions to implement your endpoints. Write detailed specifications for every function and comment your code where necessary.
4. Update the Swagger file with your API changes
5. Write at least 2 Jest tests for each endpoint

### Action Items
- [x] Implement the `/user/create` endpoint
- [x] Implement the `/user/delete` endpoint
- [x] Implement the `/user/update` endpoint
- [x] Update Swagger
- [x] Write test cases

sophwa:
Closes #11 

## Description

Allows the ability to create a user, update a current user's information, and delete an existing user from the current database. 

## Testing
Steps: 
1. Run yarn test to run the test suite. 

## Notes
This ticket adds basic functions for the manipulation of the volunteer database. Several changes are listed below: 
- A user can be added to the database when provided with a JSON body for which it will take in the necessary parameters, such as name and email. Other parameters are optional, and have a default value. 
- A current user's information can be updated in the database when provided with the user id and the updated attribute. 
- An existing user can be deleted from the database when provided with a userid.


jasozh (changes requested):
Good work!! Here are some notes:
1. `.vscode/settings.json` should be removed from the PR; I just added the `.vscode` folder to `.gitignore` so this shouldn't be a problem in the future
2. Changes to `schema.prisma` should also be removed from the PR; in general, the schema should not be modified unless it's part of the ticket. It looks like there's only formatting changes (which are needed, but I'll make those in a separate PR)
3. `/user/controllers.ts` has an extra line break on line 90
4. Tests in `users.test.ts` should follow the format in the PR Checklist Guide
5. Add a description and testing section to the PR (example shown in the PR Checklist Guide)

Also, the API just got updated so the endpoints will need to change a bit to follow the new API spec. I'll go over the new API spec in subteam as well.

PR #21, opened 2023-03-10
Merge fixes

jasozh:
## Summary
Fix several issues after dev branches were merged.

## Notes
- Delete superfluous api-spec.json
- Delete problematic prisma migration file + regenerate migrate file
- swagger autogen wasn't working because swagger.ts linked to index.ts instead of server.ts
- added swagger endpoint tags to all new endpoints
- Use Prettier to reformat all backend code
- Fix some endpoints: `/users/all` is changed to `/users` + renamed Jest tests
- Remove `forceExit` from yarn test in `package.json` because yarn test actually exits now; also added db seeding to `yarn test` so every test suite runs with a fresh batch of data (no more POST and PUT errors due to users already having been created)


akinfelami (approved):
lgtm - removed the double seeding because I already set it up in the Jest CI

Issue #20, opened 2023-03-09
Setup Firebase authentication for login and signup


jasozh:
### Goal: to setup Firebase authentication for login and signup
**Branch name:** `firebase-auth`

### Task at Hand
1. Login to [Firebase Console](https://console.firebase.google.com) with your provided account credentials.
2. Navigate to **Project settings > Service accounts** in Firebase Console and copy the Admin SDK configuration snippet. Paste the snippet in `/backend/src/server.ts` and change `var` to `const` for all variables.
3. Replace `serviceAccount` with the following object:
    <br>
    ```ts
    const serviceAccount = {
	    type: process.env.TYPE,
	    project_id: process.env.PROJECT_ID,
	    private_key_id: process.env.PRIVATE_KEY_ID,
	    private_key: process.env.PRIVATE_KEY.replace(/\\n/g, ‘\n’),
	    client_email: process.env.CLIENT_EMAIL,
	    client_id: process.env.CLIENT_ID,
	    auth_uri: process.env.AUTH_URI,
	    token_uri: process.env.TOKEN_URI,
	    auth_provider_x509_cert_url: process.env.AUTH_PROVIDER_x509_CERT_URL,
	    client_x509_cert_url: process.env.CLIENT_x509_CERT_URL,
    };
    ```
5. If you haven't already, use `yarn` to install the `firebase-admin` package. Make sure you commit `package.json` so that your dependency is included in your pull request.
6. Open `/backend/.env` and define variables for all of the `process.env.VARNAME` variables shown above. Use your provided `serviceAccountCredentials.json` file to fill in the values for each key.
7. Create the file `/backend/src/middleware/auth.ts`.
8. Read the Firebase documentation on [Verify ID Tokens](https://firebase.google.com/docs/auth/admin/verify-id-tokens) and [Control Access with Custom Claims](https://firebase.google.com/docs/auth/admin/custom-claims). Also, read this [Stack Overflow post](https://stackoverflow.com/questions/64205096/using-firebase-as-an-authenticating-middleware-in-express-js). Now, implement the following functions:
    <br>
    ```
    /** Retrieves a token from Firebase */
    getAuthToken
    
    /** Authorizes a request if a token is present, returning an error otherwise. */
    auth
    
    /** Authorizes a request if a token is present with the volunteer claim, returning an error otherwise. */
    authIfVolunteer
    
    /** Authorizes a request if a token is present with the supervisor claim, returning an error otherwise. */
[...]

PR #26, opened 2023-03-14
Firebase auth

Y1chenYao:
## Summary
1. Set up `/backend/.env` 
2. Set up `/backend/src/middleware/auth.ts` with 4 authorization functions
3. Pass authentication (only basic ones) to all existing endpoints
4. We did some experiments with creating firebase users and assigning custom claims according to the userRole ENUM, but that was aborted.

### Procedure to test
1. Make sure you have the `.env` file. See #20 or you can DM us for the content.
2. In Postman, start a POST request with the link 
    https://www.googleapis.com/identitytoolkit/v3/relyingparty/verifyPassword?key= \<API-KEY-HERE\>
    (You need the api key for the firebase service account)
3. In POST body, write
    ```
        {
          "email":"jj@prisma.io",
          "password":"12345678",
          "returnSecureToken":true
        }
    ```
5. Send and copy the idToken
6. Make a GET request with http://localhost:8000/users/
7. Past the idToken into Authorization Bearer Token and send the request

### Result
1. Without token/or with an invalid idToken, you will have 401 unauthorized error.
2. With a valid token, you will get the result just as before. For example,
image
3. With a valid token but wrong CustomClaims, for example, a valid user but not an admin to make an authIfAdmin request, you will have 401 unauthorized error.

## Notes
Several things we want to mention:
1. The scope of the ticket is very broad. We believe at first that we need to implement the full-fledged testing for the different authorization, i.e. volunteer, supervisor, and admin. However, in order to test that you will need to have ADMIN or SUPERVISOR user in the first place. We realized later that we could implement assigning custom claims but the issue is when. Do we assign the roles when the user is first created? That is our first impression when we see the 10th bullet point "Modify the endpoints GET /users/:userid and POST /users/ so that they retrieve/create a user in Firebase in addition to our database." However, now we are very confused about when we should assign the roles or create the firebase users. (Update: Akin told us that we will see that part in the frontend.)
2. The endpoints are not bug-free themselves, so it would require extra work for us if we were to make sure that POST request actually work (like creating users) and the GET request is the ticket of another pair this week. We feel that it would really alleviate our pressure if we can ensure that the endpoints actually work/exist before proceeding on integrating more features. 
3. `ts-ignores` are temporarily used. The code works but ts doesn't recogize the interface extension somehow.
4. We didn't setup the Jest testsuite so it isn't working. Maybe we should remove all the "auth" before endpoints for now?


jasozh (approved):

jasozh:
Thanks for all of the hard work this week. Tried my best to make things clear for this ticket but it was definitely more confusing than what I intended. Everything looks good and passing in the auth token works perfectly.

PR #31, opened 2023-03-20
Users sorting

### LINKED ISSUE #29: Implement sorting
### Goal: to implement sorting in our endpoints
**Branch name:** `users-sorting`

### Task at Hand
1. When a user performs an API request to one of our endpoints, they should be able to sort in ascending or descending order by various attributes.
2. Open `users/views.ts` and define a new endpoints for `GET /users/sorting`. Note that this endpoint is temporary; eventually, we would like sorting to be implemented simply in `GET /users`.
3. The endpoint will take in a single key value pair, where key `sort` has some value `sort_category:sort_order`. `sort_category` represents the column that we are sorting the data by: for example, email, first name, etc. `sort_order` is either `asc` or `desc` and represents whether we want the sorted data to be returned in ascending or descending order. For example, `GET /users/sorting?sort=email:asc` returns users ordered by email in ascending order.
4. Open `users/controllers.ts` and copy-paste the body of `getUsers` into a new function `getUsersSorted`. Edit the new function so that it supports sorting data on the following categories: email, first name, last name, and hours.
5. Update the Swagger file with your API changes.
6. Write Jest tests for your new endpoint.
7. Confirm that your code and pull request follow the [Pull Request Checklist](https://h4i.notion.site/Pull-Request-Checklist-37c67d3b6a9b488f8339f32116bfb12c).

### Action Items
- [x] Implement the endpoints
- [x] Update Swagger.
- [x] Write test cases.

bananahz:
## Summary
Closes #29 
Create endpoint for sorting users by either `email`, `hours`, `firstName`, or `lastName`

## Testing
Steps:
Run `yarn run backend` to test the endpoints on Swagger.
Run `yarn test` to run the test suite.

image
Some of the tests are on null profiles 

<insert screenshot of all tests passing> 

## Notes
We wrapped the queries in if statements for this sprint, but are looking for ways to modify our implementation with potentially a just using raw database access.


backend/tests/users.test.ts
```diff
@@ -350,6 +350,50 @@ describe("Testing /users/:userID", () => {
...
+  test("GET users sorted by firstName in descending order with null profiles", async () => {
+    const response = await request(app).get(
+      "/users/sorting?sort=firstName:desc"
+    );
+    const data = response.body;
+    expect(data[0].profile == null); //any null profiles should show up first
+    expect(data[0].profile.firstName <= data[1].profile.firstName);
+    expect(response.status).toBe(200);
+  });
+  test("GET users sorted by lastName in descending order wkith null profiles", async () => {
```
Y1chenYao, line 372: This is just a very minor formating advise. Add a line break here. :)

Y1chenYao (approved):
Very nice!!! I think the pr is pretty good!!! :)

backend/src/users/controllers.ts
```diff
@@ -563,6 +562,58 @@ const editHours = async (req: Request, res: Response) => {
...
+        include: {
+          profile: true,
+        },
+      });
+      res.status(200).json(users);
+    } else if (key == "lastName") {
+      const users = await prisma.user.findMany({
+        orderBy: {
+          profile: {
+            lastName: order == "asc" ? "asc" : "desc",
```
JJBai2003, line 603: I wonder if there is any way to remove duplicate code?

backend/src/users/controllers.ts
```diff
@@ -563,6 +562,58 @@ const editHours = async (req: Request, res: Response) => {
...
+      const users = await prisma.user.findMany({
+        orderBy: {
+          profile: {
+            firstName: order == "asc" ? "asc" : "desc",
+          },
+        },
+        include: {
+          profile: true,
+        },
+      });
```
Archit404Error, line 597: Observe that this code is precisely the same as the code on lines 600-616, you could easily pull this out into a function that simply takes in `order` as a parameter, as such:
```javascript
const findUsers = async (order : string) => {
    return prisma.user.findMany({
        orderBy: {
            profile: {
                firstName: order
            },
        },
        [...other code here]
    })
}
```

Or, with some syntactic sugar:
```javascript
const findUsers = async (order : string) =>
   prisma.user.findMany({
        orderBy: {
            profile: {
                firstName: order
            },
        },
        [...other code here]
    })
```

One thing that could be confusing is the fact that we don't `await` the async call in the function. This is because the function itself is async, so we simply `await` the overall function call via `const users = await findUsers();`

This is _precisely_ what controllers are meant to be used for (as in, functions like `findUsers()`), to ensure code re-use. This controllers file is not structured properly -- I've already mentioned this to Akin and Jason, but it's vital that you remind them to resolve this issue with a refactor.

backend/src/users/controllers.ts
```diff
@@ -563,6 +562,58 @@ const editHours = async (req: Request, res: Response) => {
...
+
+  try {
+    if (key == "email") {
+      const users = await prisma.user.findMany({
+        orderBy: [{ email: order == "asc" ? "asc" : "desc" }],
+      });
+      res.status(200).json(users);
+    } else if (key == "hours") {
+      const users = await prisma.user.findMany({
+        orderBy: [{ hours: order == "asc" ? "asc" : "desc" }],
```
Archit404Error, line 584: Rather than writing `order == "asc" ? "asc" : "desc"` multiple times throughout this code, observe that you can simply use `order` here, because order will presumably just be "asc" or "desc" so you don't need a ternary.

backend/src/users/controllers.ts
```diff
@@ -563,6 +562,58 @@ const editHours = async (req: Request, res: Response) => {
...
+          },
+        },
+        include: {
+          profile: true,
+        },
+      });
+      res.status(200).json(users);
+    }
+  } catch (error) {
+    res.status(500).json({ error: (error as Error).message });
```
Archit404Error, line 613: We should ideally have `successJson` and `errorJson` functions for these responses

Archit404Error (changes requested):
Good work, but there's some refactoring necessary here for the sake of avoiding code repetition -- let me know if you have any questions, and it is also crucial that the controller files are refactored so we _avoid_ situations like the one you've run into here

PR #32, opened 2023-03-21
Modify events

### LINKED ISSUE #27: Create endpoints for viewing and modifying events
### Goal: to create endpoints for viewing and modifying events
**Branch name:** `modify-events`

### Task at Hand
1. View the [API documentation](https://h4i.notion.site/API-Documentation-a3fe32bb2f994b02b112f68eec82d785).
2. Open `events/views.ts` and define the following endpoints:
    <br>
    ```
    GET /events/:eventid
    GET /events/:eventid/attendees
    POST /events/:eventid/attendees/:attendeeid
    DELETE /events/:eventid/attendees/:attendeeid
    PATCH /events/:eventid/status/:status
    PATCH /events/:eventid/owner/:ownerid
    PATCH /events/:eventid/attendees/:attendeeid/confirm
    ```
3. Note that all of the endpoints with `attendees` should be adding, removing, and modifying the `EventEnrollment` table in our database, not the `Events` table. We use `EventEnrollment` to store the relations between our events and our users.
4. Open `events/controllers.ts` and write functions to implement your endpoints.
5. Update the Swagger file with your API changes.
6. Write Jest tests for each endpoint.
7. Confirm that your code and pull request follow the [Pull Request Checklist](https://h4i.notion.site/Pull-Request-Checklist-37c67d3b6a9b488f8339f32116bfb12c).

### Action Items
- [ ] Implement the endpoints.
- [ ] Update Swagger.
- [ ] Write test cases.

snehar27:
## Summary

Closes #27
Creates endpoints for viewing and modifying events.

## Testing

Run `yarn run backend` to test the endpoints on Swagger.
Run `yarn test` to run the test suite.

## Notes

Endpoints have been fully implemented; however, tests for patching will have to be further inspected and rewritten.


backend/tests/events.test.ts
```diff
@@ -92,3 +93,127 @@ describe ("Testing DELETE event",() => {
...
+
+  describe ("Testing POST/events/:eventid/:attendeeid", () => {
+    test("Add attendee for existing event", async () => {
+
+      const events = await request(app).get("/events");
+      const users = await request(app).get("/users");
+      const eventID = events.body[1].id;
+      const attendeeid = users.body[1].id;
+      const response = await request(app).post("/events/" + eventID + "/"+ attendeeid);
+      expect(response.status).toBe(200);
```
akinfelami, line 140: You can be a little more clever with this test. If you add an attendee to an event, why don't you check if it was successful by seeing if that user now exists in the list of attendees for that event? 
Let me know if you have questions

backend/tests/events.test.ts
```diff
@@ -92,3 +93,127 @@ describe ("Testing DELETE event",() => {
...
+    });
+  });
+
+  describe("Testing PATCH /events/:eventid/status/:status", () => {
+    test("Update event status to active", async () => {
+      const events = await request(app).get("/events");
+      const eventid = events.body[0].id;
+      const status = "ACTIVE";
+      const response = await request(app).patch(`/events/${eventid}/status/${status}`)
+      expect(response.status).toBe(200);
```
akinfelami, line 170: You can also check if the updated event now has an active status—some tests that update user data in users tests file already do this. See that for an example

backend/tests/events.test.ts
```diff
@@ -92,3 +93,127 @@ describe ("Testing DELETE event",() => {
...
+    });
+    
+    test("Event status invalid", async () => {
+      const events = await request(app).get("/events");
+      const eventid = events.body[0].id;
+      const status = null;
+      const response = await request(app).patch("/events/" + eventid + "/status/" + status);
+      expect(response.status).toBe(500);
+    });
+  });
```
akinfelami, line 180: Another possible test - What if you use a status that is not defined as an EventStatus enum in our database schema?

backend/tests/events.test.ts
```diff
@@ -92,3 +93,127 @@ describe ("Testing DELETE event",() => {
...
+
+  describe("Testing PATCH /events/:eventid/owner/:ownerid", () => {
+    test("Change current owner", async () => {
+      const events = await request(app).get("/events");
+      const eventid = events.body[1].id;
+      const users = await request(app).get("/users");
+      const ownerid = users.body[0].id;
+      const response = await request(app).patch("/events/" + eventid + "/owner/" + ownerid); 
+      expect(response.status).toBe(200);
```
akinfelami, line 189: Again, check if the updated record now reflects the change you want - that is how you test that your function actually does what it is expected to do.

backend/tests/events.test.ts
```diff
@@ -92,3 +93,127 @@ describe ("Testing DELETE event",() => {
...
+    });
+
+  });
```
akinfelami, line 201: Isn't the same test as above? Why the repetition?

backend/tests/events.test.ts
```diff
@@ -92,3 +93,127 @@ describe ("Testing DELETE event",() => {
...
+
+  
+  describe("Testing PATCH /events/:eventid/attendees/:attendeeid/confirm", () => {
+    test("Update attendee as showed up", async () => {
+      const events = await request(app).get("/events");
+      const attendees = await request(app).get("/users");
+      const eventid = events.body[1].id;
+      const attendeeid = attendees.body[1].id;
+      const response = await request(app).patch("/events/" + eventid + "/attendees/" + attendeeid + "/confirm");
+      expect(response.status).toBe(200);
```
akinfelami, line 211: See my comments above on testing that the updated record indeed does what it is expected

backend/tests/users.test.ts
```diff
@@ -380,7 +380,6 @@ describe("Testing /users/:userID/registered", () => {
     );
     const data = GETresponse.body;
     expect(GETresponse.status).toBe(200);
-    expect(data.length).toBe(1);
```
akinfelami, line 383: To be clear, I removed this line because it is not a clever way to test in this context. We are doing all sorts of things to the seed data so we can't guarantee that the specific user will only be registered to one event

backend/tests/users.test.ts
```diff
@@ -414,22 +413,22 @@ describe("Testing /users/:userID/hours", () => {
...
+//     console.log(response.error)
+//     expect(response.status).toBe(200);
+//   });
+
+//   test("Delete invalid user", async () => {
+//     const userid = -1;
+//     const response = await request(app).delete("/users/" + userid);
+//     expect(response.status).toBe(500);
+//   });
+// });
```
akinfelami, line 431: In case you're wondering, the delete user record is commented out because it is buggy. There is a little more to deleting user records when they are constrained to other records. We will look into fixing this

backend/src/events/controllers.ts
```diff
@@ -189,6 +189,194 @@ const getPastEvents = async (req: Request, res: Response) => {
...
+        },
+        user: {
+          connect: {
+            id: req.params.attendeeid,
+          }
+        }
+      }
+    });
+
+    res.status(200).json(attendee);
```
akinfelami, line 264: Actually returning the `attendee` here is not the right approach because you are returning the entire eventEnrollment table. Look at the prisma docs or other code in the controllers file for how to use the `include` keyword

backend/src/events/controllers.ts
```diff
@@ -189,6 +189,194 @@ const getPastEvents = async (req: Request, res: Response) => {
...
+      where: {
+        id: eventID,
+      },
+    })
+
+    const attendee = await prisma.eventEnrollment.findFirst({
+      where: {
+        userId: userID,
+      }
+    })
```
akinfelami, line 291: Is this used somewhere else? I don't see why it is declared?

akinfelami (changes requested):
Hi guys. Good job on completing this ticket. I have gone through the code and added comments. Generally, you did a great job completing the endpoints and corresponding functions but I think you could have improved on the testing bit. I have added suggestions on how this can be improved. Kindly check  them out and let me know if you have any questions

backend/tests/events.test.ts
```diff
@@ -92,3 +93,133 @@ describe ("Testing DELETE event",() => {
...
+
+  describe("Testing PATCH /events/:eventid/owner/:ownerid", () => {
+    test("Change current owner", async () => {
+      const events = await request(app).get("/events");
+      const eventid = events.body[1].id;
+      const users = await request(app).get("/users");
+      const ownerid = users.body[0].id;
+      const response = await request(app).patch("/events/" + eventid + "/owner/" + ownerid); 
+      const data = response.body;
+      expect(data.ownerid).toBe("clfw74ia70000jsacl558b0ub");
```
akinfelami, line 192: You can't hardcode a an eventID like this. The test script generates a unique set of users and events everytime it is run so this will always fail

akinfelami (changes requested):

PR #36, opened 2023-03-25
jhb342 onboarding

JJBai2003:
## Summary

Completed Frontend Onboarding

## Testing

Responsive Design
Screen Shot 2023-03-25 at 3 16 31 PM

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,0 +1,9585 @@
+{
+  "name": "lfbi",
+  "version": "0.1.0",
+  "lockfileVersion": 2,
+  "requires": true,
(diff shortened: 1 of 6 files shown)


akinfelami (commented):
Great work! 
It is excellent that you can use a pure tailwind CSS grid to style divs, but MUI gives us a grid component that can be very useful and easy to use. Check it out and see how you would implement your onboarding using that. Consider different tradeoffs - development experience, speed, coding style, and more. What are your thoughts?

I also noticed to committed `package.json`. Did you do anything interesting? It appears to installed a new version of `@mui/material`?

PR #38, opened 2023-03-28
Hrz8 onboarding

bananahz:
## Summary

Front end onboarding! learned how to use Next.js and Tailwind CSS

## Notes
Small Sized Page 
image

Medium Sized Page
image

Large Page 
image

### DIFF
--- a/frontend/src/components/Button.tsx
+++ b/frontend/src/components/Button.tsx
@@ -1,0 +1,10 @@
+import React from "react";
+
+/** A simple Button component */
+const Button = () => {
+  return (
+    <button>Press me</button>
+  );
+};
+
+export default Button;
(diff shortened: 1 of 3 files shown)


akinfelami (commented):
Great work! I have two quick comments:
1. It is excellent that you can use a pure tailwind CSS grid to style divs, but MUI gives us a grid component that can be very useful and easy to use. Check it out and see how you would implement your onboarding using that. Consider different tradeoffs - development experience, speed, coding style, and more. What are your thoughts?
2. I noticed you commented out `useRouter()` . Did you find that you didn't need it? I only mention this because I believe it is something you will need to get comfortable with, along side `next/image`, `getServerSideProps`... etc using NextJs. 
Any general concerns or feedback?

PR #43, opened 2023-03-28
slw284 onboarding

sophwa:
## Summary

Learned frontend with Next.js and Tailwind.  

## Notes
Mobile: 
image

PC: 
image


akinfelami (commented):
Great work! I have two quick comments:
1. It is excellent that you can use a pure tailwind CSS grid to style divs, but MUI gives us a grid component that can be very useful and easy to use. Check it out and see how you would implement your onboarding using that. Consider different tradeoffs - development experience, speed, coding style, and more. What are your thoughts?
2. This is the frontend onboarding ticket, but you still commit changes from the backend. Are you using `git add .` for every commit? This is generally considered bad practice. The rule of thumb is to commit only files you have changed. How do you see this? in the terminal you can run `git status` to see files that have been changed and you can manually add them to the staging area. For example if you made changes to `about.tsx` and `[userid].tsx` you could run, from the root of the project - `git add frontend/src/pages/about.tsx frontend/src/pages/users/[userid].tsx` . This will ensure only both files are committed. Better still, vscode has a nice GUI tool for doing this. On the sidebar, there should be [source control](https://www.google.com/search?q=source+control+icon+vscode&rlz=1C5CHFA_enZA954ZA954&sxsrf=APwXEdfwxO8qSzjjL-OSO4H99CAPH1hqvg:1680187463709&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjZo7iH8oP-AhXRFFkFHRy8DgwQ0pQJegQIBRAC&biw=1563&bih=805&dpr=1.8#imgrc=zNJNBvcxNYdCYM) where you can see the changes, click buttons to stage, commit and push. Play around with it and let me know if you have any questions

PR #86, opened 2023-09-19
Event details frontend

### LINKED ISSUE #80: Implement Event Details
**Branch:** `event-details-frontend`

### Task at Hand
1. View the [mid-fidelity designs](https://www.figma.com/file/58CoTkFT4TgalsOyc1Patl/%F0%9F%8C%BE-Lagos-Food-Bank?type=design&node-id=1809-21467&mode=design&t=QD3MBT6azmzjOBjZ-0) for the View Events pages.
2. Implement the `EventDetails` component so that it looks like the following:
Image
4. Confirm that your code and pull request follow the [Pull Request Checklist](https://www.notion.so/h4i/Pull-Request-Checklist-910d5e6925df49569067f00f6aa0c0f3).
### Acceptance Criteria
- [ ] Users can see details for events given the parameters provided in `EventDetails`. This must be mobile friendly (see Figma)

### Notes
- Use the `IconsText` and `Chip` components that we already have (changing them if necessary)

Y1chenYao:
## Summary
- Implements `EventDetails` component. 
- Modifies `IconText` to allow custom icon/text class.
- Closes #80.

## Screenshot
Web View
image

App View
image

## Testing
Checkout `event-details-frontend`.
It is recommended to use a `text.tsx` as a page to view the component. For example,
```
import React from "react";
import EventDetails from "@/components/organisms/EventDetails";


const Test = () => {
  const image= (
  <img
    src={"https://i0.wp.com/roadmap-tech.com/wp-content/uploads/2019/04/placeholder-image.jpg?resize=800%2C800&ssl=1"}
  />)
  return (
    <EventDetails 
      title="EDUFOOD"
      location="Address, Building Name"
      datetime="02/15/2023, 9:00AM-11:00AM"
      supervisors={["Jane Doe","Jess Lee"]}
      capacity={20}
      image={image}
      tags={["In-person","EDUFOOD"]}
    />
  );
};

export default Test;
```

## Notes
1. For the input prop `image`, it would be very hard to resize without using `object-cover` css property. If passed in as a prop, we cannot style the image directly due to encapsulation and it would require extra styling before being passed in. I wonder if we can assume that image can be passed in as an URL.
2.  ~~For `IconText`, the icon color cannot be modified without modifying the original implementation. In other use cases, e.g. in `EventCancellation`, the text cannot be bolded. I wonder if we can pass in props for extending the css of icon/text.~~


frontend/src/components/atoms/IconText.tsx
```diff
@@ -12,7 +12,7 @@ const IconText = ({ icon, text }: IconTextProps) => {
   return (
     <>
       <div className="flex items-center">
-        <div className="flex-left">{icon}</div>
+        <div className={"flex-left"}>{icon}</div>
```
akinfelami, line 15: I don't believe there is a need to wrap `flex-left` in curly braces?

akinfelami (changes requested):
Great work on this! My feedback:
- Regarding the image: check out https://nextjs.org/docs/pages/api-reference/components/image and see if that will achieve what you need. I believe it should
- Yes it is okay to add props to already existing implementation if you deem fit. Just make sure to mention it in the PR so we can test the behavior properly. 

Good Jon

frontend/src/components/atoms/IconText.tsx
Y1chenYao: Oh yes! I thought that I reverted the changes but it seems that the code wasn't pushed. I will fix this along with the modifications made to IconText. :)

sophwa:
Looks good to me! The layout seems to fit what was desired in the original ticket, and it incorporates responsive design. Nice work!

snehar27 (approved):
Looks great! Follows the same design as the figma :)

akinfelami (approved):

sophwa (approved):
LGTM!

snehar27 (approved):

akinfelami (approved):
lgtm

jasozh (approved):
Looks great. Here are the things I changed:
1. The alignment between the icons and the text was not perfect in `IconsText`, I changed this by removing the `icons-center` Tailwind class. The `flex-left` classes on `icon` and `text` also didn't change the layout, so I removed those as well.
2. I removed the `iconClass` and `iconText` classes as well since it is preferable for the classes to be directly applied to the `icon` element rather than passed in as another class.
3. Because of issues with passing in `image` as a `ReactElement`, I changed the interface so that we're passing in an `image_src` tag instead. This should be fine because ultimately we will be using either image URLs or base64 strings, both of which can be expressed in a string param.

PR #100, opened 2023-09-26
User registration & Authentication Flow

akinfelami:
## Summary

This PR addresses a lot of issues/concerns with authentication. PLEASE consider the notes

## Create a user
- start the backend & frontend
- navigate to sign up page in frontend
- create a user
  - mess around with possible ways this could go wrong (try weird emails, try passwords not long enough, try a user that already exists)
- if all form entries are good, check Prisma Studio and firebase admin panel for the created user. Note that the user id is the same. Also, you are automatically signed in as soon as your account is created. Navigate to profile to see (a call to the backend) your name and details. 

## Login 
- navigate to login page
- use an already existing user to login 
- mess around with possible ways for this to go wrong - try a user that doesn't exitst, try a wrong password etc

## Auth state
- log out and select the back button
- the page should not flash any sensitive data (a loading spinner will show and redirect you back to login page)
- login using correct credentials and try navigate to the login or signup page - you should be redirected back to your profile page without mounting the login page

## Forgot password
- log out and navigate to forgot password page. 
- enter the email of the account you created earlier 
- check your email for reset email link
- click link and change password. click continue and you are redirected back to login page. 

## Profile
- update your nickname without changing the password 
- try update your password but enter a wrong old password
- try update your password but enter a new password with less than 6 characters

## Sign out
- no changes here - you should be redirected to the login page

## Notes
- This pr includes a couple of changes to jest, and package files. I upgraded the prisma version:
  - from prisma docs:
     - Node.js v16.x is reaching [end-of-life on 11 September 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol) in order to coincide with the end-of-life of OpenSSL 1.1.1. For that reason, we recommend upgrading to the current Node.js LTS, v18.x. Please note that Prisma 5 will be the last major version of Prisma to support Node.js v16.
- prettier issues - single quotes or double quotes
- with forgotpassword - firebase provides a UI for the user to reset their password. It takes some extra configuration for us to use our own webpage which is not necessary so the resetpassword page might be useless
- profile form includes a lot of changes because we are attending to two services for an update. This has to be extensively tested - it's possible I missed something
- a couple of functions in the AuthContext will become obsolete with these changes but I am keeping them just in case
- I update jest to use node version 18>
- kindly check for any trailing console.logs I might have missed


jasozh (approved):
Notes
- Reindented everything to use spaces instead of tabs
- Renamed `CustomAlert` to `Alert`
- Changed Loading from gif to MUI component
- Changed redirecting users from login pages after they're already logged in into something in `AuthContext` instead of `WelcomeTemplate`
- Redirect `/` url to `/events/view` if already logged in

frontend/src/components/organisms/SignupForm.tsx
```diff
@@ -6,127 +6,198 @@ import { useAuth } from "@/utils/AuthContext";
...
+		event?.preventDefault();
+		const { firstName, lastName, email, password } = data;
+		const emailLowerCase = email.toLowerCase();
+		const post = {
+			email: emailLowerCase,
+			profile: {
+				firstName,
+				lastName,
+			},
+			password,
```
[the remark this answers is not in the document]
jasozh: Actually wait yeah sending passwords through `POST` is not secure at all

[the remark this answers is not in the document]
jasozh: I did some more digging and this should be fine as long as we use `https` instead of `http`, which ensures that a secure encrypted SSL tunnel is established before packets are sent.

jasozh (approved):

frontend/src/components/organisms/SignupForm.tsx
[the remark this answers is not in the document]
akinfelami: Yep. As long as it is over https, it is encrypted over the wire

frontend/src/components/organisms/ProfileForm.tsx
```diff
@@ -39,8 +41,76 @@ interface ProfileFormProps {
...
   const router = useRouter();
+  const [updatePassword, updating, error] = useUpdatePassword(auth);
+  const [success, setSuccess] = React.useState<boolean>(false);
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = React.useState<string>("");
 
-  // Future testing: What if user doesn't have a nickname, does the code break?
+  const handleErrors = (errors: any) => {
+    const errorParsed = errors?.split("/")[1].slice(0, -2);
+    switch (errorParsed) {
```
danielbotros, line 51: Can we put these handleError functions in utils?

frontend/src/components/organisms/LoginForm.tsx
```diff
@@ -23,18 +40,64 @@ const LoginForm = () => {
...
+      case "auth/wrong-password":
+        return "Invalid email or password.";
+      case "auth/too-many-requests":
+        return "You have made too many requests to login. Please try again later.";
+      default:
+        return "Something went wrong.";
+    }
+  };
+
+  const LoginErrorComponent = (): JSX.Element => {
```
danielbotros, line 60: Cleaner to put this in atoms?

frontend/src/components/organisms/ProfileForm.tsx
```diff
@@ -39,8 +41,76 @@ interface ProfileFormProps {
...
+            severity="error"
+            title="Error"
+            message={handleErrors(errorMessage)}
+          />
+        ) : null}
+      </div>
+    );
+  };
+
+  const ProfileSuccessComponent = (): JSX.Element => {
```
danielbotros, line 101: I think we should either put this micro success / fail components in atoms or their own folder just to reduce the clutter

frontend/src/components/organisms/ViewEvents.tsx
```diff
@@ -13,34 +13,35 @@ import Link from "next/link";
...
-        mainAction={"rsvp"}
-        dropdownActions={["cancel rsvp"]}
-        title={"snehar"}
-        location={"jameson"}
-        datetime={"today lol"}
-      />
-    </CardList>
-  );
+  return <></>;
+  // return (
```
danielbotros, line 17: Try not to merge this into main while commented out (remove, change, keep, etc...) 🙏

danielbotros (approved):

frontend/src/components/organisms/ProfileForm.tsx
jasozh: There does seem to be a lot of duplication with the `handleError` functions; we probably should be able to abstract it into a helper method since the Firebase error codes are all standardized. Will look into this

frontend/src/components/organisms/ProfileForm.tsx
jasozh: Definitely there does seem to be a lot of duplication but putting them in atoms isn't the right solution; atoms should be for generic reusable components with no business logic tied to them, and in this case, all of the micro components are just wrappers around the existing `Alert` atom. Will look into different ways to refactor this

frontend/src/components/organisms/ViewEvents.tsx
jasozh: Was commented out for deployment purposes but will be uncommented once a different PR is merged (which fixes event cards)

frontend/src/components/organisms/ProfileForm.tsx
akinfelami: I agree but will take some care to duplicate it because the different functionalities have slightly different error codes/messages

frontend/src/components/organisms/ProfileForm.tsx
akinfelami: Maybe could be in utils or similar

PR #115, opened 2023-10-10
Refactor POST /:eventid/attendees

jasozh:

### DIFF
--- a/backend/api-spec.json
+++ b/backend/api-spec.json
@@ -501,9 +501,7 @@
           }
         ],
         "responses": {}
-      }
-    },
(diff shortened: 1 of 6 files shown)


akinfelami (approved):
Looks good to me!

I think it makes sense to pass the user ids in the body and not params. Appreciate the effort to refactor this

jasozh:
> Side note: Might be a good idea (just like what you have done here) to incrementally refactor the backend before we roll out the tickets with respect to the endpoints that will be used. This way, we don't accumulate bugs

Yep that's the general idea, I'll be making more backend refactoring PRs like this as the sprints progress

PR #116, opened 2023-10-10
Refactor volunteer registration flow

jasozh:


akinfelami (commented):
Looks good to me! 

The one concern I have is having two `register.tsx` files. I understand what you are trying to achieve but it could be confusing in the long run

jasozh:
I agree this could be a concern but I couldn't think of a better solution. We can keep this for now then and refactor it when we come up with a better alternative.

PR #117, opened 2023-10-18
Events users backend

### LINKED ISSUE #114: Implement searching, sorting, and pagination (backend)
**Branch:** `events-users-backend`

### Task at Hand

1. Open `/backend/src/users/views.ts` and `/backend/src/events/views.ts`.
2. Implement the acceptance criteria.
3. Confirm that your code and pull request follow the [Pull Request Checklist](https://www.notion.so/h4i/Pull-Request-Checklist-910d5e6925df49569067f00f6aa0c0f3).

### Acceptance Criteria

- [x] `GET /users` is an endpoint that combines searching, sorting and pagination. For example, `GET /api/users?firstName="Bobby"&lastName="Fischer"&sort=email:asc&limit=100&after=xjf48992` should be a valid API endpoint that gets the first 100 users with name "Bobby Fischer" sorted by email in ascending order.
- [x] `GET /events` is an endpoint that combines sorting and pagination. For example, `GET /api/events/upcoming=true?sort=location:desc&limit=20&after=xjf48992` should be a valid API endpoint that gets the first 20 upcoming events sorted by location in descending order.
- [x] All of the sorting and pagination from `GET /users` and `GET /events` should be applicable to any API endpoint that returns similar functionality. The following endpoints should be implemented: `GET /users/:userid/created`, `GET /users/:userid/registered`, `GET /events/:eventid/attendees`.

### Notes

- View the [API documentation](https://www.notion.so/h4i/API-Documentation-956584fab90649f889094520153317f4) to understand how we are doing pagination, sorting, and searching.
- Each of these functions should already be implemented individually. Take a look at the controllers file for both `users` and `events` to see existing implementations. This ticket should be **combining** the functionality from these functions; you shouldn't have to write it anew.

bananahz:
## Summary
- Closes #114 
- Consolidates `GET /users` search, sort, and pagination functionalities. 
- Consolidates `GET /events` sort and pagination functionalities. 

`GET /users/:userid/created` -> `GET /events?ownerid=[userid]`
(note: `GET /users/:userid/created` doesn't return a list of events currently, but assumed functionality is that it should) 

`GET /users/:userid/registered` -> `GET /events?userid=[userid]`
Note how for this specific query, it includes `attendees` as part of the response 
image

`GET /events/:eventid/attendees `->  `GET /users?:eventid=[eventid]`
image

## Testing
- Checkout `events-users-backend`
- Request should be in the format of `GET /api/users?` and `GET /api/events?` 
- Sample requests: 
http://localhost:8000/events?ownerid=[ownerid]&after=[id]
http://localhost:8000/events?ownerid=[ownerid]&upcoming=true&after=[id]
http://localhost:8000/events?upcoming=true&sort=location:desc&limit=20&after=[id]
http://localhost:8000/users?firstName=Alice&sort=email:asc&limit=100&after=[id]
> Note: IDs will be different each time. We recommend testing manually alongside with Prisma Studio (for filtering, sorting, etc.)

## Testing on Pagination
Previously we were worried that Prisma cannot take a column that is not unique. Now setting the id to be a tuple `(search, id)` guarantees its uniqueness.

### User Query

Filter by volunteer, sorted according to email, take the first 10 (default) after cursor `felipa34@yahoo.com, id`.

Expectation: we expect the 10th one to be `marilou71@hotmail.com`.
image

Result: Indeed the last one is `marilou71@hotmail.com`. The other ones are manually checked.
image

### Event Query 

Sort events in descend order, take the first 3 after cursor `soluta quos beatae, id`.

Expectation:
image

Result: 
image

## Notes/Questions 
- We didn't remove any existing functions (search, sorting, pagination), should we? 
- Cursors must be passed in, otherwise an empty list is returned. Can we assume that cursor is always provided?

### DIFF
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -45,77 +45,94 @@ const deleteEvent = async (eventID: string) => {
 };
 
 /**
- * Search, Sort and Pagination for Events
+ * Search, sort and pagination for events
  * [option] corresponds to the columns in the Event table.
  * Search supports multiple queries
  * The following options are supported:
- * @param upcoming if the event is upcoming then true, else false
- * @param sort possible sort columns: name, location
- * @param ownerid all events that were created by ownerid
- * @param userid all events such that userid is registered 
- * @param req: Request paramters to get query used for search
- * @returns  promise with list of all events where [option] is [value]
-
+ * @param filter are the filter params passed in
+ * @param sort are sort params passed in
+ * @param pagination are the pagination params passed in
+ * @returns promise with list of all events where [option] is [value]
  */
-const getEvents = async (req: Request) => {
-  const query = req.query;
-  let upcoming = query.upcoming;
-  let whereDict: { [key: string]: any } = {};
-  let includeDict: { [key: string]: any } = {};
-  if (upcoming) {
-    if (upcoming == "true") {
-      const dateTime = new Date();
-      whereDict["startDate"] = {
-        gt: dateTime,
-      };
-    }
+const getEvents = async (
+  filter: {
+    upcoming?: string;
+    ownerId?: string;
+    userId?: string;
+  },
+  sort: {
+    key: string;
+    order: Prisma.SortOrder;
+  },
+  pagination: {
+    after: string;
+    limit: string;
   }
-  const sortQuery = query.sort as string;
+) => {
+  /* SORTING */
+
+  // Handles GET /events?sort=location:desc
   const defaultCursor = { id: "asc" };
-  const querySplit = sortQuery ? sortQuery.split(":") : ["default", "asc"];
-  const key: string = querySplit[0];
-  const order = querySplit[1] as Prisma.SortOrder;
   const sortDict: { [key: string]: any } = {
-    default: { id: order },
-    name: [{ name: order }, defaultCursor],
-    location: [{ location: order }, defaultCursor],
+    default: { id: sort.order },
+    name: [{ name: sort.order }, defaultCursor],
+    location: [{ location: sort.order }, defaultCursor],
   };
 
-  const ownerId = query.ownerid;
-  if (ownerId) {
-    whereDict["ownerId"] = ownerId;
-  }
-  const userId = query.userid;
-  if (userId) {
-    whereDict["attendees"] = {
-      some: {
-        userId: userId,
-      },
-    };
-  }
+  /* PAGINATION */
+
+  // Handles GET /events?limit=20&after=asdf
   let cursor = undefined;
   let skip = undefined;
-  if (query.after) {
+  if (pagination.after) {
     cursor = {
-      id: query.after as string,
+      id: pagination.after as string,
     };
     skip = 1;
   }
   let take = undefined;
-
-  if (query.limit) {
-    take = parseInt(query.limit as string);
+  if (pagination.limit) {
+    take = parseInt(pagination.limit as string);
   } else {
     // default take is 10
     take = 10;
   }
+
+  /* FILTERING */
+
+  let whereDict: { [key: string]: any } = {};
+  let includeDict: { [key: string]: any } = {};
+
+  // GET /events?upcoming=true
+  if (filter.upcoming === "true") {
+    const dateTime = new Date();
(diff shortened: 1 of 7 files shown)


backend/src/events/controllers.ts
```diff
@@ -39,13 +45,68 @@ const deleteEvent = async (eventID: string) => {
...
+  return prisma.event.findMany({
+    where: {
+      AND: [whereDict],
+    },
+    include: includeDict,
+    orderBy: sortDict[key],
+    take: query.limit ? parseInt(query.limit as string) : 10,
+    cursor: {
+      id: query.after ? (query.after as string) : "",
+    },
```
akinfelami, line 107: I have questions on how cursor works.  If you take look at this link https://www.prisma.io/docs/concepts/components/prisma-client/pagination#:~:text=%7D)-,Cursor%2Dbased%20pagination,-Cursor%2Dbased%20pagination
In the code, I think yours is missing the part where they bookmark the location in the result set. Then how does cursor based pagination help us because the cursor is never updated

akinfelami: Making cursor required will not be ideal. For example, I copied the ID of the user just after Alice in prisma studio in this query and I didn't get any result. 
`users?firstName=Alice&sort=email:asc&limit=100&after=clo0s0ax40001ccb712nntgbf`

akinfelami (changes requested):
Hi guys! Great work on this so far!

I don't have any specific code nitpicks yet but I do have concerns functionality wise. Mostly on `cursor`. I added comments in the review which apply to both `events` and `users`.

Also, what did `GET /events/:eventid/attendees` change to? I see your PR description includes a comment that 
Screenshot 2023-10-22 at 22 36 18
I don't think this is good approach. I don't think it makes sense to want to get the attendees for an event and you pass in a query with a userid. I think it should be something like this

`GET /events/:eventid/attendees` -> GET /events?eventId=[eventid]. This seems like a more appropriate approach to get attendees for an event. 

Let me know what you think

backend/src/events/controllers.ts
Y1chenYao: > I have questions on how cursor works. If you take look at this link https://www.prisma.io/docs/concepts/components/prisma-client/pagination#:~:text=%7D)-,Cursor%2Dbased%20pagination,-Cursor%2Dbased%20pagination In the code, I think yours is missing the part where they bookmark the location in the result set. Then how does cursor based pagination help us because the cursor is never updated

I thought that it was used only for demontration, since `myCursor` is only a variable and not necessarily part of the implementation. I could however, return the response containing the `cursor id`. 

> Making cursor required will not be ideal. For example, I copied the ID of the user just after Alice in prisma studio in this query and I didn't get any result.
users?firstName=Alice&sort=email:asc&limit=100&after=clo0s0ax40001ccb712nntgbf

I will fix this right now.

Y1chenYao: Hi Akin, all changes have been applied. Hopefully it is to your expectation. :)
- For retrieving the first page, `cursor id` can be omitted. Next cursor is incorporated in the response.
image

- For future retrievals, `cursor id` is needed.
image

akinfelami: Hi, Yichen. This looks really clever. I will test and revert

Y1chenYao: Hi Akin, I looked into the database and currently we only have 1 enrollment. Did you add more attendees manually for this?

akinfelami: Hey, yes! I did. I modified my seed scripts which populated my database with a lot user enrollment. Can you copy the entire seed file from here - https://github.com/cornellh4i/lagos-volunteers/blob/event-registration-fullstack/backend/prisma/seed.ts 

You will also need to update `backend/src/utils/script.ts` https://github.com/cornellh4i/lagos-volunteers/blob/event-registration-fullstack/backend/src/utils/script.ts

akinfelami (approved):

DannyThorne01 (approved):
Great Work!!

louisvalencia (approved):
Looks good! I would suggest comments for the long functions like getEvents because it could be a little hard to follow

jasozh (approved):

PR #132, opened 2023-11-18
Table pagination frontend 2

DannyThorne01:
## Summary

Clicking on the pagination buttons in the MUI Data Grid performs a function/API call instead of initiating the built-in pagination functionality.

## Testing
You'll only need the local host to test this.
On the website navigate to the `Manage Users` tab, then click the pagination buttons to the lower right of the table. You should be able to switch between different users. 

image
image

## Notes
I made some changes to the `Table` prop: Notably:
image
 I added `dataSetLength` and `initialID` so that I can have some starting ID to begin the pagination process and a limit to end pagination.
`Next Function` is needed to actually implement pagination and move from one page to the next.
With these new additions it would mean that any component calling the `Table` component would probably need two additional functions. One to obtain the `dataSetLength` and `initialID` and another to act as `Next Function`. This is shown in the `ManageUsers` component.

### DIFF
--- a/frontend/src/components/molecules/Table.tsx
+++ b/frontend/src/components/molecules/Table.tsx
@@ -1,72 +1,65 @@
 import React from "react";
 import Box from "@mui/material/Box";
-import {
-  DataGrid,
-  GridColDef,
-  GridColumnHeaderParams,
-  GridValueGetterParams,
-  MuiEvent,
-  gridPageCountSelector,
-  GridPagination,
-  useGridApiContext,
-  useGridSelector,
-} from "@mui/x-data-grid";
-
-import MuiPagination from "@mui/material/Pagination";
-
-interface TablePaginationProps {
-  page: number;
-  onPageChange: (event: React.MouseEvent | null, page: number) => void;
-}
-function Pagination({ page, onPageChange }: TablePaginationProps) {
-  const apiRef = useGridApiContext();
-  const pageCount = useGridSelector(apiRef, gridPageCountSelector);
-  return (
-    <MuiPagination
-      color="primary"
-      count={pageCount}
-      page={page + 1}
-      onChange={(event, newPage) => {
-        onPageChange(event as any, newPage - 1);
-      }}
-    />
-  );
-}
-
-function CustomPagination(props: any) {
-  return <GridPagination ActionsComponent={Pagination} {...props} />;
-}
+import { DataGrid, GridColDef, GridPaginationModel } from "@mui/x-data-grid";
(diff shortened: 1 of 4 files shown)


frontend/src/components/molecules/Table.tsx
```diff
@@ -1,42 +1,99 @@
-import React from "react";
-import {
-  DataGrid,
-  GridColDef,
-  GridColumnHeaderParams,
-  GridValueGetterParams,
-  MuiEvent,
-} from "@mui/x-data-grid";
+import React, { useEffect } from "react";
+import Box from "@mui/material/Box";
```
bananahz, line 2: remove unused import

frontend/src/components/molecules/Table.tsx
```diff
@@ -1,42 +1,99 @@
...
+          paginationModel={paginationModel}
+          onPaginationModelChange={handlePaginationModelChange}
+          pageSizeOptions={[]}
+          paginationMode="server"
+          disableColumnMenu
+        />
+      )}
+    </div>
   );
 };
```
bananahz, line 98: I'm assuming this PR is a WIP, there are a bunch of console logs which should be removed once code is finalized

frontend/src/components/organisms/ManageUsers.tsx
```diff
@@ -225,9 +113,92 @@ const Active = () => {
...
+        method: "GET",
+        headers: {
+          Authorization: `Bearer ${userToken}`,
+        },
+      });
+      if (response.ok) {
+        const data = await response.json();
+
+        const length = data["data"].length;
+        const first_user_id = data["data"][0]["id"];
```
bananahz, line 138: tiny nitpick, for consistency, maybe change this variable name to use camelcase

bananahz (changes requested):
Not sure if code is finalized yet, left some comments regarding code style, good work so far!

Y1chenYao (changes requested):
The table looks really nice!
Clicking on the `view profile` would not lead me to correct page, since in `ManageUsers.tsx` it has the dummy page of `href="/users/asdf/manage`, totally understanding if this is not part of the scope. :)

akinfelami (commented):
I think this is good to go! I would say I didn't inspect the logic too deeply but it works as intended. Also, the request doesn't filter for Blacklisted users. We might need to update the PR Hannah and Yichen worked on to include this filter! Besides that it looks good to go!

jasozh (approved):

PR #134, opened 2023-12-21
Add fetch wrapper function for API calls

jasozh:
## Summary

Create API helpers to reduce code duplication from API fetch calls.

For example, this

```
  const fetchUrl = `${url}/users/search/?email=${user?.email}`;
  const userToken = await auth.currentUser?.getIdToken();
  const response = await fetch(fetchUrl, {
    method: "GET",
    headers: {
      Authorization: `Bearer ${userToken}`,
    },
  });
  const data = await response.json();
```

becomes

```
const { data } = await api.get(`/users/search/?email=${user?.email}`);
```


frontend/src/utils/api.ts
akinfelami, line 1: This is a cool idea! But what do you think about having only one function called `callApi` or something but the function will take `URL`, `method` (default to `GET`) and `body` (optional of course) and in the function we just pass in the body if it is provided. I think that might simplify things even more because instead of having 4 functions we have 1. Do you get what I mean?

akinfelami: That makes sense. We could probably still reduce code duplication by having the a `handleRequest` helper just like the `handleResponse` which takes in the arguments I mentioned above. So in the individual `get` `post` and so you use the handleRequest helper instead

jasozh: Good idea, just implemented it.

akinfelami (approved):
I didn't run it locally but it lgtm!

PR #135, opened 2023-12-22
Update dependencies

jasozh:
## Summary

Update package dependencies specified in the `yarn.lock` file. Ensures that future invocations of `yarn install` uses the updated package versions.

### DIFF
--- a/backend/yarn.lock
+++ b/backend/yarn.lock
@@ -10,58 +10,58 @@
     "@jridgewell/gen-mapping" "^0.3.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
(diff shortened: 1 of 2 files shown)


akinfelami (approved):

jasozh:
Dependency updates are creating some issues with JSX elements not being recognized. Need to debug before we can merge.

akinfelami:
> Dependency updates are creating some issues with JSX elements not being recognized. Need to debug before we can merge.

Can you share more? How can I reproduce?

jasozh:
Check out the branch, do a `yarn install` with the new lock files and present, and open any React file. At least for me, every single React component has a red squiggly line that says it's not a valid type or something like that. Doing a `yarn install` on the old lock files resolves the issue.

akinfelami:
That's interesting. I am not facing the same issue. Everything looks okay on my local

PR #136, opened 2023-12-23
Refactor atoms, molecules, and organisms

jasozh:
## Summary

Significant overhaul of various, atoms, and molecules. Details are specified below:
- `Alert` now conforms to the high fidelity designs
- `Button` is rewritten to remove unnecessary props and conforms to the high fidelity designs, with four variants: primary, secondary, tertiary, error
- `Chip` is refactored
- `DatePicker`, `TimePicker` is rewritten to allow `react-hook-form` to utilize it in forms. Also refactored.
- `LocationPicker` is completely overhauled; `react-google-autocomplete` is replaced with pure MUI since the autocomplete menu is significantly prettier and uses pure MUI components. Making MUI `Autocomplete` work with `react-hook-form` is impossible however, so a kinda-hack was used to set the form value directly (see the `form` prop)
- `SearchBar`, `MultilineTextField` and `TextField` are refactored to remove unnecessary props, clean up the code, and make them look like the high fis
- Completely new `Select` component is created
- Completely new `Snackbar` component is created to create floating notifications. All alerts throughout the website now use floating notifications rather than the old alerts
- `TextCopy` conforms to high-fis
- `AppBar` and `NavBar` is now rewritten to conform to high-fis and fix the placement of the Log Out button in style terms
- `Avatar` now conforms to high-fis
- `TabContainer` now conforms to high-fis
- `Table` is not really changed but columns/rows throughout the website are adjusted to look prettier
- `EventCard` now changes button color depending on the event date (if in the future, main action icon is white instead of green)
- `EventForm` now uses `react-hook-form` for all date and time values. Old validation is subsequently removed where needed.
- Welcome pages now conform to high fis. Same with View Events and Manage Attendees (in only some areas, didn't fully implement the high fi design)
- Inter is now imported and used as the website font; an MUI theme is created as well to use for the `primary` color
- `react-google-autocomplete` is removed and a single (very small) dependency `autosuggest-highlight` is added to make the new LocationPicker work

## Notes

Example of the new Login form
image


akinfelami:
Okay so this PR and the other refactor one are pretty chunky. I don't think I can be as thorough in reviewing them. I'm running into some bugs here and there when I try to run both locally. I'm guessing that moving things around might have overlooked some edge cases. I think you can merge both branches into a new branch and let us test extensively from there

Here's an example of a bug I ran into
Screenshot 2023-12-23 at 16 59 56

jasozh:
I think `refactor-templates` should stay a second branch because the changes there are completely visual, so that branch shouldn't break any functionality. Meanwhile this branch makes a lot of changes to everything.

There's a lot of errors but I think the majority of those are also on the main branch. For example, the error you showed just now was something I was also able to reproduce on `main`, see below:

image

I think the website is generally still pretty buggy right now after I merged all of the fullstack tickets, but those fixes are out of scope for this PR (since it's chunky enough as is) so I avoided touching them.

One new error I can think of is that when floating notifications are sent, they kinda glitch out and reappear/disappear at once. This will be a wontfix for now because I think we should try to overhaul our entire error state management in another PR (maybe have an error toggle `open, setOpen` instead of display/hide a React component) so I didn't want to touch it in this PR.

akinfelami:
Okay, this bug specifically wasn't there when I tested the PR Dan worked on. Something from must have broken when we merged everything. I think the error state management can be improved with the react-query I mentioned. I haven't seen what the "floating notifications" look like though. However, if I try to attempt a overhaul on the fetching, error and loading states in this PR, it is going to be ginormous, but still, I am skeptical about merging this code into main since we know it is broken

jasozh:
If you look at the change log for `ManageUserProfile`, the only changes I added to the file was to add floating notifications via the new `Snackbar` component. The functionality might be broken right now but it's not caused by any changes from this PR. I have noticed that the bug is only present for some userids and not others, so that's probably why I didn't catch it while merging the fullstack tickets.

I don't think it's a good idea to overhaul fetching + error state in this PR for the reasons you mentioned. If you start a clean branch from `main`, there'll also be a lot of issues with trying to merge the two branches back together.

frontend/package.json
```diff
@@ -22,6 +22,7 @@
     "@types/node": "18.11.19",
     "@types/react": "18.0.27",
     "@types/react-dom": "18.0.10",
+    "autosuggest-highlight": "^3.3.4",
```
akinfelami, line 25: what are you using this package for?

jasozh: For the new location picker, I copy-pasted this snippet of code from the [MUI documentation](https://mui.com/material-ui/react-autocomplete/#google-maps-place). `autosuggest-highlight` was a required dependency and it seems like it's needed to have highlighted text in the autosuggest menu.

akinfelami (approved):

PR #140, opened 2024-01-12
Event registration hifis

jasozh:
## Summary

Implement hi-fi designs for the Event Registration page. Note: new pages are frontend only changes and existing pages are not removed yet. This is just a frontend skeleton.

image

## Testing

Create the following `test.tsx` file:
```
import React from "react";
import EventRegisterCard from "@/components/organisms/EventRegisterCard";
import ViewEventDetails from "@/components/organisms/ViewEventDetails";

const Test = () => {
  return (
    <ViewEventDetails />
    // <div className="bg-black p-5">
    //   <EventRegisterCard action="cancel confirmation" />;
    // </div>
  );
};

export default Test;
```

Try changing the `action` prop on `EventRegisterCard`

### DIFF
--- a/frontend/src/components/organisms/ViewEventDetails.tsx
+++ b/frontend/src/components/organisms/ViewEventDetails.tsx
@@ -47,6 +47,7 @@ const ViewEventDetails = () => {
           <div className="font-semibold text-xl">About the event</div>
           <Divider />
           <div className="mt-5"></div>
+
           <div>FOOD BANK COMMUNITY OUTREACH</div>
           <div>Kindly find below the details of the outreach:</div>
           <div className="mt-5"></div>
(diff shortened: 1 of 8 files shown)


frontend/src/components/organisms/ViewEventDetails.tsx
akinfelami, line 1: I see the About the event section having multiple divs and spacing and the likes. I think we need to develop around the fact that the event description will be given as a giant blob of text with no sections. What do you think?

akinfelami (commented):
lgtm

frontend/src/components/organisms/ViewEventDetails.tsx
jasozh: The spacing divs are temporary, but I agree we'll need a better solution eventually. I think we might want to use `react-quill` for the event description as well, similar to what we used for the FAQ page

akinfelami: Recently came accross https://platejs.org/ which might be worth looking into for rich text editing

akinfelami (approved):

frontend/src/components/organisms/ViewEventDetails.tsx
jasozh: Seems promising, and the live demo definitely looks a lot more sleek + modern than `react-quill`. Will need to play around with it a bit first though, and there seems to be a lot of dependencies.

PR #142, opened 2024-01-13
Implement event card hifis

jasozh:
## Summary

Implement frontend skeleton for the event card changes.

image

image

## Testing

Use the following `test.tsx` file:
```js
import React from "react";
import EventCardNew from "@/components/organisms/EventCardNew";
import DefaultTemplate from "@/components/templates/DefaultTemplate";

const Test = () => {
  return (
    <DefaultTemplate>
      <EventCardNew />
      <div className="mt-5" />
      <EventCardNew />
      <div className="mt-5" />
      <EventCardNew />
    </DefaultTemplate>
  );
};

export default Test;

```

## Notes

Frontend is actually so hard.


akinfelami (approved):
lgtm

PR #143, opened 2024-01-14
Implement dropzone

jasozh:
## Summary

Implement a dropzone component that allows uploading image files. Uploaded files also have an image preview. Includes basic input validation and requires a `setState` function to be passed in to handle errors.

image
image

## Testing

Use the following `test.tsx` file:
```js
import React, { useState } from "react";
import Dropzone from "@/components/atoms/Dropzone";
import DefaultTemplate from "@/components/templates/DefaultTemplate";

const Test = () => {
  const [error, setError] = useState("");
  return (
    <DefaultTemplate>
      <div className="p-10 bg-white">
        <button onClick={() => console.log(error)}>Test</button>
        <Dropzone setError={setError} />
      </div>
    </DefaultTemplate>
  );
};

export default Test;
```

## Notes
Dropzone based on the Flowbite dropzone: https://flowbite.com/docs/forms/file-input/#dropzone.

How do we stop people from uploading viruses? Basic file validation is implemented but not sure if this is enough...

### DIFF
--- a/frontend/src/components/atoms/Dropzone.tsx
+++ b/frontend/src/components/atoms/Dropzone.tsx
@@ -1,0 +1,82 @@
+import React, { useState } from "react";
+
+interface DropzoneProps {
+  setError: React.Dispatch<React.SetStateAction<string>>;
+}
+
+/**
+ * Dropzone component that allows uploading files. Requires a setState to be
+ * passed in to handle file upload errors.
+ */
+const Dropzone = ({ setError }: DropzoneProps) => {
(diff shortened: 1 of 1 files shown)


akinfelami (approved):
Looks good to me!

Good callout on the how to stop people from uploading viruses. I think the enforcing file format might be enough ?

jasozh:
Images can still contain viruses, so just enforcing file format isn't enough. But in the worst case scenario, we can just disable uploading a custom avatar during signup and only enable uploading images for events (which only supervisors can do).

PR #144, opened 2024-01-16
Adjust all fetching & state management to use react-query

akinfelami:
## Summary

This is a quite chunky PR. I have changed a bunch of things over a few weeks so I might still be tidying up stuff as you review this PR. The biggest introduction is [react-query]( https://tanstack.com/query/latest/docs/react/overview ). It is a very useful library that gives a lot of functionality. We get more deterministic loading, error and success states to manage the app better. In my opinion, it is not as straightforward as traditional useEffects and useStates (development-wise) but the functionality we get out of it is worth it. I'm not sure I can get to all I changed in the PR but I will highlight most of it. Also, I can't guarantee that I didn't introduce any new bugs (some might be very niche) but I did try to address all bugs I found since I started working on this

PS: I already started working on this before your last (46 file change) PR was merged into main so I just merged it into my branch hence you will see commits from that PR!

## Testing

Honestly just click through the site and see if anything breaks lol. 

## Notes

- I'm working on a notion page to come up with this PR. This will include most of the changes I made and minor issues/features that I think are worth looking at
- Most of the code additions you will find come from pages that require pagination. It was a headache! but I think they all work perfectly now.
- A cool feature we get from react-query is caching. So when you navigate to the user profile page for example and navigate somewhere else but click the back button to go to the profile page, there is no loading state because it shows the cached data. 
- Another cool feature is that we don't need to reload like we used to for registering changes to the user. React-query will invalidate the cached query when it changes. For example, if you change some data in a user profile. The changes immediately show up on the screen without the need to reload the page. Same occurrence for registering and canceling events!
- The entire site also includes some UX we didn't consider before -> like when a user has no upcoming events; we don't want to show a blank page. Also, if the loading fails for some reason, we need to show an error page. I improvised for most of these but you get the idea
- Finally please see the comments in my review for some more context


frontend/src/components/organisms/ViewEvents.tsx
```diff
@@ -188,53 +201,262 @@ const PastEvents = ({ eventDetails }: EventCardProps) => {
...
+      startDate: event["startDate"],
+      endDate: event["endDate"],
+      role: userid === event["ownerId"] ? "Supervisor" : "Volunteer",
+      hours: computeHours(event["startDate"], event["endDate"]),
+    });
+  });
+
+  // Handle Past Events
+  // We need to separate the query for past events registered for and past events supervised
+  // because we need to handle pagination differently for each (state is different)
```
akinfelami, line 279: note this (probably why this page is chunky

frontend/src/pages/events/[eventid]/cancel.tsx
```diff
@@ -8,70 +8,63 @@ import { fetchUserIdFromDatabase, formatDateTimeRange } from "@/utils/helpers";
...
-    updateEventDetails();
-  }, []);
+  const userHasCanceledAttendance =
+    eventAttendance && eventAttendance["canceled"];
+
+  if (!eventAttendance) {
+    router.push(`/events/${eventid}/register`);
+  }
+
+  if (isLoading) return <Loading />;
```
Comment (akinfelami) [inline frontend/src/pages/events/[eventid]/cancel.tsx:60]: note this is how we get the loading spinner before the data arrives

akinfelami (commented):
Added comments to stuff I think are worth noting!

jasozh:
Very nice work! This will take me a while to review so I'm planning on doing it file by file. I'll highlight anything I change along the way.

backend/src/events/controllers.ts
```diff
@@ -100,12 +100,20 @@ const getEvents = async (
...
+    case "upcoming":
+      whereDict["startDate"] = {
+        gt: dateTime,
+      };
+      break;
+    case "past":
+      whereDict["endDate"] = {
+        lt: dateTime,
+      };
+      break;
```
jasozh, line 116: Replaced `GET /events?upcoming=true` and `GET /events?upcoming=false` with `GET /events?date=upcoming` and `GET /events?date=past` for cleaner API syntax

backend/src/users/views.ts
```diff
@@ -18,43 +18,37 @@ userRouter.post(
   NoAuth as RequestHandler,
   async (req: Request, res: Response) => {
     // #swagger.tags = ['Users']
+    // if you're seeing this, no you didn't
```
jasozh, line 21: lmao

backend/src/users/views.ts
```diff
@@ -18,43 +18,50 @@ userRouter.post(
...
+
+      // If local user doesn't exist, throw error
+      if (!user) {
+        throw Error("Failed to create local user");
+      }
+
+      // If test environment, return user without firebase auth
+      else if (process.env.NODE_ENV === "test") {
+        return res.status(201).send({ success: true, data: user });
+      }
```
jasozh, line 40: Old changes were breaking the test cases because nothing was being returned if the env variable was set to `test`. Added the return back and cleaned up the code a bit.

frontend/src/components/atoms/Button.tsx
```diff
@@ -9,6 +9,7 @@ interface ButtonProps {
   size?: "small" | "medium";
   loading?: boolean;
   [key: string]: any;
+  type?: "button" | "submit" | "reset";
```
jasozh, line 12: Removed this because it's not actually needed since it is already included in the `[key: string]: any` prop. Removing it means our Button interface is cleaner since we don't need to include any props that are implicitly included with the traditional `button`. e.g. `onClick` is not in our interface but we can still pass it in as usual

frontend/src/components/molecules/Loading.tsx
```diff
@@ -1,12 +1,17 @@
...
 function Loading() {
   return (
-    <div className="h-screen bg-white">
-      <div className="flex justify-center items-center h-full">
+    <div>
+      <Backdrop
+        sx={{ color: "#fff", zIndex: (theme) => theme.zIndex.drawer + 1 }}
+        invisible
+        open
+      >
```
jasozh, line 12: Use MUI Backdrop so that the loading spinner is always centered in the middle of the screen (although tbf our loading times are so fast now idk if people will ever see this component)

frontend/src/components/organisms/EventCard.tsx
```diff
@@ -13,11 +13,11 @@ import { Action } from "@/utils/types";
 import { formatDateTimeRange } from "@/utils/helpers";
 
 interface EventCardProps {
-  eventid: string;
+  eventid: string | undefined;
```
jasozh, line 16: Replaced with `eventid?; string;` because it's functionally equivalent and cleaner.

backend/src/users/views.ts
akinfelami: Good catch!

frontend/src/components/molecules/Loading.tsx
akinfelami: I think we would need to account for varying internet speeds so it's possible we see faster load times because our internet is great but bad internet speed will leave the user with a blank screen (poor UX)

frontend/src/components/organisms/NavBar.tsx
```diff
@@ -13,8 +13,7 @@ const NavBar = () => {
   const handleSignOut = async () => {
     try {
       await signOutUser();
-      // clear cache!!
-      queryClient.clear();
+      queryClient.clear(); // Clear cache for react query
```
akinfelami, line 16: The reason why I did this is relevant. I actually discovered this (bug?) accidentally
- For pagination, each page has its own unique query key using the page number
- I set the staleTime to Infinity so that the data is not updated with background re-fetches (react-query does this). This way, we can check previous and next pages seamlessly
- now when a user signs out and another user signs in they will see the data of the other user that was cached before the actual data is updated so we need to clear the entire cache when the session changes

frontend/src/components/molecules/Loading.tsx
jasozh: Gotcha. Probably also helps that both our server and client are on localhost, which would explain the blazing fast loading times.

frontend/src/utils/api.ts
```diff
@@ -99,7 +99,7 @@ const handleResponse = async (response: Response) => {
     } else {
       data = await response.blob();
     }
-    return { response: response, data: data };
+    return Promise.resolve({ response: response, data: data });
```
jasozh, line 102: What's the difference between wrapping a value in `Promise.resolve` and returning a value directly?

akinfelami: I don't remember exactly why I did this now. I guess I thought if the else case was rejected then the if condition should probably resolve.

frontend/src/components/organisms/LoginForm.tsx
```diff
@@ -94,89 +93,70 @@ const LoginForm = () => {
...
+        onClose={() => setNotifOpen(false)}
+      >
+        Error: {handleErrors(signInErrors?.code)}
+      </Snackbar>
+
+      {/* Login form */}
+      <div className="space-y-4">
+        <img src="/lfbi_logo.png" className="w-24" />
+        <form onSubmit={handleSubmit(handleLogin)} className="space-y-4">
+          <div className="font-bold text-3xl"> Log In </div>
```
jasozh, line 121: Moved `LoginErrorComponent` down here because having it as its own component was breaking the `Snackbar` close animation. Also changed `Snackbar` to disable clickaway (so it's harder to accidentally dismiss the notification)

jasozh: Also didn't put `Snackbar` inside the div with the form because doing so creates a visual glitch whenever the Snackbar is activated.

jasozh: Also refactored the notifications everywhere else, including `ProfileForm`, `EventForm`, etc.

frontend/src/components/atoms/DatePicker.tsx
```diff
@@ -37,7 +37,7 @@ const CustomDatePicker = forwardRef(
               textField: { fullWidth: true, error: error !== "" },
             }}
             ref={ref}
-            format="MM/DD/YYYY"
+            format="DD/MM/YYYY"
```
jasozh, line 40: Changed to dd/mm/yyyy

frontend/src/components/organisms/EventRegisterForm.tsx
```diff
@@ -1,85 +1,81 @@
...
+  });
 
   return (
     <div>
-      <h2>Terms and Conditions</h2>
-      <p>
+      <Box sx={{ textAlign: "center" }}>
+        <h2>Terms and Conditions</h2>
+      </Box>
+      <Box sx={{ marginBottom: 3 }}>
```
jasozh, line 51: Some styling changes to the modal

frontend/src/components/organisms/LoginForm.tsx
akinfelami: sounds good

frontend/src/components/organisms/Error.tsx
```diff
@@ -0,0 +1,16 @@
...
+    <div className="flex justify-center items-center h-screen">
+      <div className="text-center">
+        Aw! An error occurred while fetching your data :(
+        <p>Please try again</p>
+      </div>
+    </div>
+  );
+};
+
+export default Error;
```
jasozh, line 16: We'll make it prettier later

frontend/src/utils/helpers.ts
```diff
@@ -96,7 +95,20 @@ export const convertToISO = (inputTime: string, inputDate: string) => {
...
+ * @param endTime is the end time
+ * @returns the hours
+ */
+export const eventHours = (endTime: string, startTime: string) => {
+  const end = new Date(endTime);
+  const start = new Date(startTime);
+  const diff = end.getTime() - start.getTime();
+  const hours = diff / (1000 * 3600);
+  return hours;
+};
```
jasozh, line 114: Moved to `helpers.ts` from `ManageUserProfile`

frontend/src/components/organisms/ViewEvents.tsx
jasozh, line 1: Moved most of the code into the subcomponents `<UpcomingEvents />` and `<PastEvents />` for code clarity.

jasozh (approved):
Wow that was a lot of changes. I've done a bunch of random stuff so feel free to take a look through all of my changes and see if there's anything you disagree with. A lot of code moved around but the vast majority of it is just adding comments + moving lines up and down, and every other change I've tried to document as I was reviewing.

Great work, especially `ViewEvents` which scared me quite a bit. We've changed about 6,000 lines of code since the start of winter break which is crazy.

akinfelami:
I was monitoring your changes and they all seem fine to me. Good work adding comments and tidying things up. Two final changes from me:
- Pushed a last fix that caused an error to pop up even on successful login. Turns out the `setNotifOpen` was not in the catch block lol. 
- Pushed a schema change to add a `phoneNumber` column for user profile and also used [Referential actions](https://www.prisma.io/docs/orm/prisma-schema/data-model/relations/referential-actions#referential-action-defaults) for Prisma to properly delete stuff. You will notice the delete user controller now only requires deleting the user table and Prisma handles the deletion of other tables that have a relationship with it.

Looks like there is one test failing in the CI but seems to be passing locally. I will investigate and lyk when it is good to go.

Ultimately, great work this winter break! We definitely achieved a lot. I will still share the doc I said I was working on. 


6000 line changes though :)
Someone Cooked Here Meme

PR #145, opened 2024-02-08
Update dependencies

jasozh:
## Summary

Update dependency versions in `package.json`. Did some preliminary testing to make sure nothing was broken.


akinfelami (commented):
Can we leave this open for now ? until we merge the other large PR. I don't assume anything will be broken, but for good measure, I say we merge this after testing with that PR

jasozh:
I think it should be fine because we can always revert back to the old yarn.lock and package.json versions if an issue does come up. I wasn't able to find any issues but feel free to test it whenever you're free. I think reading through the changelog might be a bit too much, but I understand the concern.

frontend/package.json
```diff
@@ -11,36 +11,35 @@
...
-    "firebase": "^9.18.0",
-    "next": "13.1.6",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
+    "date-fns": "^3.3.1",
+    "dayjs": "^1.11.10",
+    "eslint": "^8.56.0",
+    "eslint-config-next": "^14.1.0",
+    "firebase": "^10.8.0",
+    "next": "^14.1.0",
```
akinfelami, line 31: https://nextjs.org/blog/next-14#other-changes

akinfelami (approved):

akinfelami (changes requested):
hold off to fix  merge conflicts

akinfelami (approved):
good to go!

PR #152, opened 2024-02-17
Revert enums back to uppercase

akinfelami:
## Summary

changing to sentence case enums was shortsighted. Apologies for the headache and here's my chance at redemption. 

## Testing

I have clicked through the website. I tried changing user roles, statuses and the likes. Also tried create a new event to see if it still works. Everything works fine. For good measure, I looked into the 2 commits, I used to initially change it to sentence case and tried to make sure I undid everything. ran all the tests locally. Everything works fine so far, including seeding.


jasozh (approved):
looks good

PR #167, opened 2024-03-02
Verify Email Fullstack

DiegolMarques:
## Summary

When the user signs up for the first time, they are asked now to verify their email with a unique link, which verifies them in the firebase and the prisma. They are not able to log in to the application until they are verified. Additionally, there is a new verification page which the user is sent to. 

## Testing

Manual Testing:

Tested Login:
- Non-verfiied
- Verfiied
- Non-user

Tested Signup
- Non-verfiied
- Verfiied
- Non-user

## Notes

We added a function in the helpers file to update a verified status in the prisma database.

### DIFF
--- a/frontend/src/components/organisms/SignupForm.tsx
+++ b/frontend/src/components/organisms/SignupForm.tsx
@@ -12,6 +12,7 @@ import { useSignInWithEmailAndPassword } from "react-firebase-hooks/auth";
 import Snackbar from "../atoms/Snackbar";
 import { api } from "@/utils/api";
 import { useMutation } from "@tanstack/react-query";
+import { sendEmailVerification } from "firebase/auth";
 
 type FormValues = {
   firstName: string;
--- a/frontend/src/components/organisms/SignupForm.tsx
+++ b/frontend/src/components/organisms/SignupForm.tsx
@@ -8,11 +8,13 @@ import { BASE_URL } from "@/utils/constants";
 import { useForm, SubmitHandler } from "react-hook-form";
 import Alert from "../atoms/Alert";
 import { useRouter } from "next/router";
-import { useSignInWithEmailAndPassword } from "react-firebase-hooks/auth";
+import {
+  useSignInWithEmailAndPassword,
+  useSendEmailVerification,
+} from "react-firebase-hooks/auth";
 import Snackbar from "../atoms/Snackbar";
 import { api } from "@/utils/api";
 import { useMutation } from "@tanstack/react-query";
-import { sendEmailVerification } from "firebase/auth";
 
 type FormValues = {
   firstName: string;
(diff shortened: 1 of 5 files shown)


frontend/src/pages/verify.tsx
```diff
@@ -0,0 +1,101 @@
...
+  //         // If user is not signed in, redirect to login page
+  //         router.push("/login");
+  //       }
+  //     } catch (error) {
+  //       console.error("Error checking user verification:", error);
+  //     }
+  //   };
+
+  //   checkUserVerification();
+  // }, []);
```
akinfelami, line 54: I have commented out this code because I pushed some new code to AuthContext.tsx. The code I pushed will check if a user is verified or not and route them accordingly. This means there will be no need to do that check here again. 

Your new changes will simply be functions that will be called when an event happens. For now, let us leave firebase as the source of truth for email verification flag

frontend/src/pages/verify.tsx
```diff
@@ -0,0 +1,101 @@
...
+            If you haven't received the email within a few minutes, please check
+            your spam folder.
+          </p>
+          <Button
+            onClick={handleResendEmail}
+            variety="primary"
+            className="w-full mb-4"
+          >
+            Resend Verification Email
+          </Button>
```
akinfelami, line 91: We need to give the user some feedback for clicking this button. First, when you use the useMutation hook, you will get an `isPending` state which you can pass into the loading prop of this button. This means that the user will be able to see when something is loading. Additionally, you can add  a `SnackBar` to indicate that the email was successfully sent or something went wrong. Please see similar implementations in `ManageUserProfileNew.tsx`

frontend/src/utils/helpers.ts
akinfelami, line 1: I imagine the change in this file will no longer be necessary after you consider other comments

akinfelami (changes requested):
Great work guys! Awesome work getting this done!

I have added some comments to your file changes, please take a look and don't hesitate to ask any clarifying questions. Thanks!

frontend/src/components/organisms/SignupForm.tsx
```diff
@@ -96,10 +101,15 @@ const SignupForm = () => {
       // Log in
       const { email, password } = data;
       const signedInUser = await signInWithEmailAndPassword(email, password);
+      const emailSent = await sendEmailVerification(); // Send email verification
```
akinfelami, line 104: Hmmm so I think I only received the email after I clicked on the resend email verification in the verify page. I didn't automatically get it after signin up. Are u able to reproduce?

hehub000: We weren't able to reproduce this error; when we go to the verify page we automatically got the email.

akinfelami (commented):
This PR is currently blocked pending some new changes to reset password and the firebase console

> The firebase template for sending any emails only uses one link, which can't be the same for email verification, password reset, etc. (It seems like firebase assumes we have all this functionality in one file). We propose in the future changing our multiple file system for auth into one like firebase suggests.

PR #171, opened 2024-03-09
Reset Password Fullstack

DiegolMarques:
## Summary

Added reset password functionality to the reset page. Changed authcontext to include all different reset password links to be apart of the public paths. Changed firebase email template for use with localhost. Implemented Ticket. 

## Testing

Test was done manually. Edge cases covered:
If passwords don't match
If user is not logged in
All different reset route links
etc

## Notes

The firebase template for sending any emails only uses one link, which can't be the same for email verification, password reset, etc. (It seems like firebase assumes we have all this functionality in one file). We propose in the future changing our multiple file system for auth into one like firebase suggests.


frontend/src/components/organisms/ResetPasswordForm.tsx
```diff
@@ -54,7 +90,7 @@ const ResetPassword = () => {
         />
       </div>
       <div>
-        <Button>Reset Password</Button>
+        <Button type="submit">Reset Password</Button>
```
akinfelami, line 93: the`useUpdatePassword` hook from react-firebase-hooks should provide some isLoading Prop that we can pass into the loading prop of this button. So when the user clicks the button, they get some feedback that something is happening. Here's an excerpt from LoginForm

```js
  /** Firebase hooks */
  const [
    signInWithEmailAndPassword,
    signedInUser,
    signInLoading,
    signInErrors,
  ] = useSignInWithEmailAndPassword(auth);

...

<Button
      loading={signInLoading}
      disabled={signInLoading}
      type="submit">
            Log in
  </Button>
```

frontend/src/utils/AuthContext.tsx
```diff
@@ -100,24 +100,29 @@ export const AuthProvider = ({ children }: AuthProviderProps) => {
...
+    "/password/reset/",
   ];
 
   // Paths that can be accessed freely when not logged in
   const publicPaths = [
     "/login",
     "/signup",
     "/password/forgot",
-    "/password/reset/*",
+    "/password/reset",
```
akinfelami, line 111: why remove asteriks? I thought the asteriks is supposed to match all the links that have /password/reset?

frontend/src/utils/AuthContext.tsx
```diff
@@ -100,24 +100,29 @@ export const AuthProvider = ({ children }: AuthProviderProps) => {
     "/login",
     "/signup",
     "/password/forgot",
-    "/password/reset/*",
+    "/password/reset/",
```
akinfelami, line 103: why remove asteriks? I thought the asteriks is supposed to match all the links that have /password/reset?

akinfelami (changes requested):
Hi guys!

Amazing work on this. You guys have really done well finding the relevant documentation and edge cases for this. 

Like we discussed, this is currently blocked from merging due to changes in verify user email address. But, there are still comments worth addressing that I have added to the PR. Kindly take a look and let me know if there are any questions. Looking forward to discussing!

frontend/src/utils/AuthContext.tsx
DiegolMarques: We found out that the asteriks does nothing it just takes in the string as it is, so the asteriks will never happen unless the route is named '*'

frontend/src/pages/password/reset.tsx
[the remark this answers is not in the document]
hehub000: There was something going on with AuthContext that would check /password/reset/[link] multiple times and eventually take the user back to the login page. We couldn't figure out what the problem was but changing the reset page to it's own file rather than a folder and just checking for that in AuthContext didn't give us the same issue.

frontend/src/components/organisms/ResetPasswordForm.tsx
```diff
@@ -3,24 +3,60 @@ import Button from "../atoms/Button";
...
+    name: string,
+    url: string = window.location.href
+  ): string | null => {
+    name = name.replace(/[[\]]/g, "\\$&");
+    const regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
+      results = regex.exec(url);
+    if (!results) return null;
+    if (!results[2]) return "";
+    return decodeURIComponent(results[2].replace(/\+/g, " "));
+  };
```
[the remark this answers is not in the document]
DiegolMarques: This is a standard convention for parsing URL query parameters, which extracts the value of a specific parameter from a given URL. It starts by replacing square brackets in the parameter name to ensure literal matching. Then, it constructs a pattern to match the parameter name followed by an equal sign, capturing the parameter's value until it encounters an ampersand, hash sign, or the end of the string. This expression accommodates various URL formats and is commonly used in web development for extracting query parameters from URLs.

PR #184, opened 2024-03-15
Refactor login and signup error handling

jasozh:
## Summary

Previous Firebase login and signup error handling did not make effective use of hooks. This PR changes it so that error handling makes use of `signInErrors` provided by `react-firebase-hooks` and `error` provided by `tanstack-query`. This should make error handling more streamlined than before since we're using the error objects provided instead of trying to do `try { catch }` blocks ourselves.

## Testing

Test login and signup.

## Notes

I think more error handling should probably make use of `isError` and `error` provided by `tanstack-query` later on.


akinfelami (changes requested):
a few comments!

akinfelami (approved):
lgtm

PR #188, opened 2024-03-15
Add markdown editor

jasozh:
## Summary

Replace `react-quill` with a simple Markdown editor for rich text editing. Benefits include a standardized way of rendering Markdown as HTML without us needing to render raw HTML generated by `react-quill`, and also because it would make the actual data in the backend much easier to read since we can just store in plaintext.

For example, nested bullets completely fail to render in `react-quill`

image
image

## Testing

Create a new `pages/test.tsx` file and paste:
```typescript
import React, { useState, useCallback } from "react";
import "easymde/dist/easymde.min.css";
import Markdown from "react-markdown";

// Import SimpleMDE
// cannot just do import SimpleMDE from "react-simplemde-editor";
// it will break with "navigator not defined"
// see https://github.com/dabit3/next.js-amplify-workshop/issues/21
import dynamic from "next/dynamic";
const SimpleMdeEditor = dynamic(() => import("react-simplemde-editor"), {
  ssr: false,
});

export const Test = () => {
  const [value, setValue] = useState("Initial value");

  const onChange = useCallback((value: string) => {
    setValue(value);
  }, []);

  return (
    <>
      <button onClick={() => console.log(value)}>click</button>
      <SimpleMdeEditor value={value} onChange={onChange} />
      <Markdown>{value}</Markdown>
    </>
  );
};

export default Test;
```

## Notes
`
- Will use in `About.tsx` and `EventForm.tsx

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -25,6 +25,7 @@
     "autosuggest-highlight": "^3.3.4",
     "date-fns": "^3.3.1",
     "dayjs": "^1.11.10",
+    "easymde": "^2.18.0",
     "eslint": "^8.56.0",
     "eslint-config-next": "^14.1.0",
     "firebase": "^10.8.0",
@@ -34,7 +35,9 @@
     "react-firebase-hooks": "^5.1.1",
     "react-hook-form": "^7.50.1",
     "react-html-parser": "^2.0.2",
+    "react-markdown": "^9.0.1",
     "react-quill": "^2.0.0",
+    "react-simplemde-editor": "^5.2.0",
     "typescript": "^5.3.3"
   },
   "devDependencies": {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,9 +19,6 @@
     "@mui/x-data-grid": "^6.19.3",
     "@mui/x-date-pickers": "^6.19.3",
     "@tanstack/react-query": "^5.18.1",
-    "@types/node": "^20.11.16",
-    "@types/react": "^18.2.55",
-    "@types/react-dom": "^18.2.19",
     "autosuggest-highlight": "^3.3.4",
     "date-fns": "^3.3.1",
     "dayjs": "^1.11.10",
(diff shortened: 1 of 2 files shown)


akinfelami (commented):
I like this, it works great!

But we are working under the assumption that our partners know how to write markdown. We might want to operate under the assumption that they have only used text editing tools like google docs and Microsoft word

jasozh:
I agree but I've looked through a lot of different rich text editor and all of them are massive headaches to work with. `react-quill` is by far the easiest to work with but even then it is incredibly jank in actually rendering the text output. I've seen some straggling GitHub threads on converting `react-quill` output to markdown, but those are also very buggy. I think it should be intuitive enough because the toolbar still includes buttons for bolding, italicizing, etc.

akinfelami (commented):
I think it is alright to use this, although I will point out that there is a way to get the pure text in react-quill (without the HTML tags) but here the markdown syntax is included in the text. 

Also, if we intend to use this for description of events, will it be an issue in the future since event descriptions will be stored as markdown? 

Ah , and I move the type declaration dependencies to devDependencies so that it is not bundled with the build for production environment, which means slimmer build files

jasozh:
> I think it is alright to use this, although I will point out that there is a way to get the pure text in react-quill (without the HTML tags) but here the markdown syntax is included in the text.

I think it's fine if markdown is included since at most it'll just be `**bold**` or `*italic*`.

> Also, if we intend to use this for description of events, will it be an issue in the future since event descriptions will be stored as markdown?

Nope, it should make event descriptions just as easy to render as the FAQ page. We'll replace the event description text box with another markdown editor component.

> Ah , and I move the type declaration dependencies to devDependencies so that it is not bundled with the build for production environment, which means slimmer build files

Okay, sounds good.

jasozh:
I also found another markdown editor which might be better suited for our usecase called [MDXEditor](https://mdxeditor.dev/). It's WYSIWYG and also fully markdown compatible so checks all our boxes (cause I really want us to be able to render it in HTML with `react-markdown`, but I messed around with it and could not figure out how to get it working with Next.js server-side rendering.

If you're able to get it working, feel free to add it to this branch. Otherwise I would recommend just merging this for now since I do think either way we want to end up working with markdown.

PR #189, opened 2024-03-15
Add node cron

jasozh:
## Summary

Add node cron to run scheduled jobs periodically in the server. Needed for next week's sprint.

## Testing

Start the server and check the console every minute.


akinfelami (approved):
lgtm!

Changed from require to import , thought it would be nice to be consistent

PR #192, opened 2024-03-17
Got global sorting on email column. Still have to fix cursor issue

### LINKED ISSUE #187: Table sort (fullstack)
**Branch:** `table-sort-fullstack`

### Task at Hand

1. Implement your code in `Table.tsx` and any component that uses a `Table` component
2. Confirm that your code and pull request follow the [Pull Request Checklist](https://www.notion.so/h4i/Pull-Request-Checklist-910d5e6925df49569067f00f6aa0c0f3).

### Acceptance Criteria

- [ ] Pressing the sort arrows on a table should perform a global sort. Currently, only the results in the current page are sorted, but instead, pressing the sort arrows should initiate a new API call that sorts the entire dataset.

### Notes

- See the [MUI documentation](https://v4.mui.com/components/data-grid/sorting/#server-side-sorting) for implementing server-side sorting. Also, reference our existing code where we implement server-side pagination since it should look very similar.

dvala041:
Closes #187

## Summary

<li> Added properties to the table component for server-side global sorting
<li> modified useEffect() in ManageUser.tsx to perform the server-side sorting
<li> still have to fix issue with cursor being used more than once when refetching

## Testing

<li> tested by using console.log()

## Notes
<li> Once again still some changes we have to make for this to work correctly. 
<lI> Also the other files that use tables must be adjusted as they are throwing errors

### DIFF
--- a/backend/src/users/controllers.ts
+++ b/backend/src/users/controllers.ts
@@ -203,7 +203,6 @@ const getUsers = async (
     ? queryResult[take - 1]
     : queryResult[queryResult.length - 1];
   const myCursor = lastPostInResults ? lastPostInResults.id : undefined;
-  // console.log(queryResult);
   return { result: queryResult, cursor: myCursor, totalItems: totalRecords };
 };
 
--- a/backend/src/users/controllers.ts
+++ b/backend/src/users/controllers.ts
@@ -208,8 +208,14 @@ const getUsers = async (
   const lastPostInResults = take
     ? queryResult[take - 1]
     : queryResult[queryResult.length - 1];
-  const myCursor = lastPostInResults ? lastPostInResults.id : undefined;
-  return { result: queryResult, cursor: myCursor, totalItems: totalRecords };
+  const nextCursor = lastPostInResults ? lastPostInResults.id : undefined;
+  const prevCursor = queryResult[0] ? queryResult[0].id : undefined;
+  return {
+    result: queryResult,
+    nextCursor: nextCursor,
+    prevCursor: prevCursor,
+    totalItems: totalRecords,
+  };
 };
 
 /**
--- a/backend/src/users/controllers.ts
+++ b/backend/src/users/controllers.ts
@@ -203,6 +203,7 @@ const getUsers = async (
     ? queryResult[take - 1]
     : queryResult[queryResult.length - 1];
   const myCursor = lastPostInResults ? lastPostInResults.id : undefined;
+  // console.log(queryResult);
   return { result: queryResult, cursor: myCursor, totalItems: totalRecords };
 };
 
--- a/backend/src/users/controllers.ts
+++ b/backend/src/users/controllers.ts
@@ -126,7 +126,7 @@ const getUsers = async (
   // Handles GET /users?limit=20&after=asdf
   let cursor = undefined;
   let skip = undefined;
-  if (pagination.after) {
+  if (pagination.after != "") {
     cursor = {
       id: pagination.after as string,
     };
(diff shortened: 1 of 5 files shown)


backend/src/users/controllers.ts
```diff
@@ -203,6 +203,7 @@ const getUsers = async (
     ? queryResult[take - 1]
     : queryResult[queryResult.length - 1];
   const myCursor = lastPostInResults ? lastPostInResults.id : undefined;
+  // console.log(queryResult);
```
akinfelami, line 206: remove comments.

frontend/yarn.lock
akinfelami, line 1: remove this file

akinfelami (changes requested):
Folks, great work on this!

I have added some code and I believe the sorting is now functional. Please test this to confirm. 

Additionally, please remove commented out code. but leave necessary comments. Also remove any unnecessary logs. 

Finally address the comments I have included in the PR. Great job!

dvala041:
### Fixes

<li> included sortingOrder property in datagrid component in tables.tsx

### Bugs

1. When staleTime is infinity in the Tanstack query, the next pages don't change when we change sort order
2. Even when we comment out the staletime line there is another bug:
a. When we're on say page 4 and hit the sort button, the sort mode changes but page 0 doesn't change. If we traverse to the next pages the data is all wrong. 
b. Akin noted that the data is being fetched again every time we change pages which isn't intended behavior (should only be fetched when sorting model changes)

backend/src/users/controllers.ts
```diff
@@ -126,7 +126,7 @@ const getUsers = async (
   // Handles GET /users?limit=20&after=asdf
   let cursor = undefined;
   let skip = undefined;
-  if (pagination.after) {
+  if (pagination.after != "") {
     cursor = {
       id: pagination.after as string,
     };
```
akinfelami, line 132: this is to address the case of passing an empty cursor, if an empty cursor is passed, the backend should be able to ignore it.

backend/src/users/controllers.ts
```diff
@@ -186,24 +186,36 @@ const getUsers = async (
...
+    queryResult = await prisma.user.findMany({
+      ...queryOptions,
+      cursor,
+    });
+  } else {
+    queryResult = await prisma.user.findMany({
+      ...queryOptions,
+    });
+  }
+
```
akinfelami, line 207: more on passing empty cursor

backend/src/users/controllers.ts
```diff
@@ -186,24 +186,36 @@ const getUsers = async (
...
-  const myCursor = lastPostInResults ? lastPostInResults.id : undefined;
-  return { result: queryResult, cursor: myCursor, totalItems: totalRecords };
+  const nextCursor = lastPostInResults ? lastPostInResults.id : undefined;
+  const prevCursor = queryResult[0] ? queryResult[0].id : undefined;
+  return {
+    result: queryResult,
+    nextCursor: nextCursor,
+    prevCursor: prevCursor,
+    totalItems: totalRecords,
+  };
```
akinfelami, line 218: New design here:

Based on cursor pagination. NextCursor is the id of the last user returned for this batch. We will use this to fetch the next batch (basically users after this id in the db). PrevCursor is the id of the first user in the batch. So if we need to fetch a previous page, we just take the users before that id

akinfelami: This also means that all components that used to consume this endpoint will break because what this returns has changed

frontend/src/components/organisms/ManageUsers.tsx
```diff
@@ -147,76 +163,154 @@ const Active = ({
...
+    { field: "firstName", sort: "asc" },
+  ]);
+
+  /** Function to fetch a batch of users with respect to the current pagination
+   * and sorting states.
+   * @param cursor is the cursor to fetch the next batch of users
+   * @param prev is a boolean to determine if the function is being called to fetch the previous page
+   * Note: If the prev boolean is true, we pass a negative value to the limit parameter to fetch
+   * data before the current cursor. This is effectively the previous page.
+   * @returns the current page of users (with respect to the current pagination and sorting states)
```
akinfelami, line 185: I tried to put comments on some of these functions so they are not too difficult to follow, the order I put them is a good way to understand what they are doing

akinfelami (commented):
new pagination model and sorting model that seems to work. Deprecates all that prefetching we used to do. Lmk what you think

jasozh:
I like this. So basically we have a `prevCursor` and a `nextCursor` which defines the start and end of the batch, and we use the `prevCursor` to grab the previous page and `nextCursor` to grab the next page. Tested it out and it looks like sorting actually works which is pretty awesome.

PR #193, opened 2024-03-17
About page fullstack new

### LINKED ISSUE #125: Implement About Page (fullstack)
**Branch:** `about-page-fullstack`

### Task at Hand

1. View the [mid-fidelity designs](https://www.figma.com/file/58CoTkFT4TgalsOyc1Patl/%F0%9F%8C%BE-Lagos-Food-Bank?type=design&node-id=1809-21467&mode=design&t=QD3MBT6azmzjOBjZ-0) for the About page.
2. Open the `About` component.
3. Implement the component with the acceptance criteria below.
4. Confirm that your code and pull request follow the [Pull Request Checklist](https://www.notion.so/h4i/Pull-Request-Checklist-910d5e6925df49569067f00f6aa0c0f3).

### Acceptance Criteria

- [ ] Admins are able to edit the About page in a text box and save their changes.
- [ ] Regular users only see the standard About page.

### Notes

- Take a look at the backend functions and the API documentation if you need reminders on what backend functions are available

owenchen04:
## Summary

Changed About.tsx to align with hi-fi designs. Added new About table to database so that edits made to about page content in the frontend is reflected in the backend.

## Testing

We tested by seeding the About table with the default page content, then made changes using the `Edit Text` feature. We checked Prisma studio to make sure that the changes in the frontend were reflected in the backend. Api routes were tested in Postman.

## Notes

- Since our last week's ticket (role specific pages) hasn't been merged yet, we weren't able to complete the functionality on only allowing admins to see and interact with the `Edit Text` button. However, after that is merged, the logic to incorporate this feature should be very straightforward.
- For some reason, registering the about api routes under `users/views.ts` was giving us issues. As a temporary fix, we've implemented the about backend functions directly inside `server.ts`, which seemed to work fine.
- We couldn't find a fix for the whitespace bug for the time being.
- Future ticket will implement uploading images.
- Closes #125


frontend/src/components/organisms/About.tsx
```diff
@@ -1,108 +1,177 @@
...
+        newContent: newContent,
+      });
+      return data;
+    },
+    retry: false,
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ["about"] });
+    },
+    onError: (e: Error) => {
+      console.log(e.message);
```
akinfelami, line 97: You need to implement `SnackBar` for success and error feedback to user

frontend/src/components/organisms/About.tsx
```diff
@@ -1,108 +1,177 @@
...
+  const {
+    data: aboutPageDetailsQuery,
+    isPending: aboutPageFetchPending,
+    isError,
+  } = useQuery({
+    queryKey: ["about"],
+    queryFn: async () => {
+      const { response, data } = await api.get("/about");
+      // console.log(data);
+      setValue(data.content);
```
akinfelami, line 68: the beauty of tanstack query is that you don't need to manage states like this, so you actually don't need this. The `data` holds all the information you need,

frontend/src/components/organisms/About.tsx
```diff
@@ -1,108 +1,177 @@
...
+  const handleEditClose = () => {
+    setEditMode(false);
+    setValue(lastValueRef.current);
+  };
   const handleModalClick = () => {
     setEditMode(false);
-    handleClose();
+    handleConfirmClose();
+    lastValueRef.current = value;
+    changeAboutContent({ newContent: value });
```
akinfelami, line 117: rename this function to something that better indicates what you are trying to do.

akinfelami: Also the function is async and should be awaited

frontend/src/components/organisms/About.tsx
```diff
@@ -1,108 +1,177 @@
...
+  const handleEditOpen = () => setEditMode(true);
+  const handleEditClose = () => {
+    setEditMode(false);
+    setValue(lastValueRef.current);
+  };
   const handleModalClick = () => {
     setEditMode(false);
-    handleClose();
+    handleConfirmClose();
+    lastValueRef.current = value;
```
akinfelami, line 116: not sure if this is necessary, why do you need it?

akinfelami (changes requested):
Hi guys!

great work on this one!

I think the issue you were facing is that you didn't realize all the endpoint in the users/views are wrapped in the `/users` middleware, so you probably needed to call the endpoint like this `/users/pageid`. Please revisit this and move the function appropriately

Also, there are few other comments that need to be addressed. Let me know if there are any questions

frontend/src/components/organisms/About.tsx
owenchen04: hi Akin; I believe we use the useState hook to manage the data in the frontend so that we can store the edits that the user makes in ReactQuill and then call the updateAboutPage on that data so that the content is synced appropriately between the front- and backends. I'm not sure how else I'd be able to keep track of the user's edits without using useState; unless I'm missing something?

`        <ReactQuill
          theme="snow"
          value={value}
          onChange={setValue}
          readOnly={false}
        />`

frontend/src/components/organisms/About.tsx
tunile943: Because before this, when you edit you will edit the about content directly, so when you close it the changes will still be saved. I added this to ensure that when you don't want to save the edit and want to revert to the old edit, the old value is still there for us to revert back to.

akinfelami:
@owenchen04 I think the issue is that it should have been await api.get("/users/about"); not await api.get("users/about");. Notice the slash before users

Please move the endpoints into a new view/controllers module. And please remember to add Authentication middleware, as the current implementation does not use any authentication

owenchen04:
> @owenchen04 I think the issue is that it should have been await api.get("/users/about"); not await api.get("users/about");. Notice the slash before users
> 
> Please move the endpoints into a new view/controllers module. And please remember to add Authentication middleware, as the current implementation does not use any authentication

will do!

backend/src/users/views.ts
akinfelami, line 1: remove commented out code

akinfelami (changes requested):
one last thing!

akinfelami (approved):
Fixed merge conflicts in last commit.

jasozh (approved):

PR #198, opened 2024-03-18
Manage attendees fullstack

hehub000:
## Summary

Implemented the functionality according to the ticket description, with some details left unimplemented.

## Testing

Manually tested using multiple users and verified against prisma database.

## Notes

The rows don't display any status initially (the select box is blank) and the page has to be refreshed for a user to be moved to the correct tab, but otherwise the functionality is there. We decided to PR without fully fixing those details because we didn't want thie PR to come any later.


frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -364,26 +398,34 @@ const ManageAttendees = ({}: ManageAttendeesProps) => {
       // TODO: Double check endpoint
       // It currently returns list of ALL users, not just attendees for a specific event
       const { data } = await api.get(
-        `/users?eventid=${eventid}&limit=${paginationModel.pageSize}`
+        `/events/${eventid}`
```
akinfelami: you don't want to change this. There was a slight bug with it but I have pushed a change now. the correct query will be `await api.get(`/users?eventId=${eventid}&limit=${paginationModel.pageSize}`);`. You can log this in the console to see that it gives you all the information you need for a user's enrollment. No need to make another endpoint to get the user's enrollment status

frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -346,7 +380,7 @@ const ModalBody = ({
 };
 
 /** A ManageAttendees component */
-const ManageAttendees = ({}: ManageAttendeesProps) => {
+const ManageAttendees = ({ }: ManageAttendeesProps) => {
   const router = useRouter();
   const eventid = router.query.eventid as string;
```
akinfelami, line 385: the eventid is already here btw

frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -119,6 +151,8 @@ const AttendeesTable = ({
     console.log(value);
   };
 
+  const filteredRows = rows.filter((attendee: attendeeData) => attendee.status === status);
```
akinfelami: I think the filtering should happen before you pass the prop into the table

backend/src/events/controllers.ts
```diff
@@ -280,6 +281,31 @@ const getAttendees = async (eventID: string, userID: string) => {
...
+  return await prisma.eventEnrollment.findUnique({
+    where: {
+      userId_eventId: {
+        userId: userID,
+        eventId: eventID,
+      },
+    },
+  });
+}
+
```
akinfelami, line 308: you will not be needing this with the new changes I pushed. Please ask any questions if you have on why

backend/src/events/views.ts
```diff
@@ -106,6 +106,22 @@ eventRouter.post("/:eventid/attendees", async (req: Request, res: Response) => {
...
+});
+
+eventRouter.get("/:eventid/users/:userid", async (req: Request, res: Response) => {
+  // #swagger.tags = ['Events']
+  attempt(res, 200, () =>
+    eventController.getEnrollmentStatus(req.params.eventid, req.params.userid)
+  );
+});
+
+
```
akinfelami, line 124: remember to remove this

frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -20,17 +20,21 @@ import { api } from "@/utils/api";
 import { useRouter } from "next/router";
 import Loading from "../molecules/Loading";
 import Card from "../molecules/Card";
+import { User } from "firebase/auth";
+import { userAgentFromString } from "next/server";
```
akinfelami, line 24: idk what this is or if it is being used

frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -44,7 +48,34 @@ type FormValues = {
...
+  try {
+    // Make a PUT request to update the user
+    const response = await api.put(`/events/${eventId}/users/${userId}`, {
+      status: newValue // Only update the status field
+    }); console.log("success")
+  } catch (error) {
+    console.error("Error updating user status:", error);
+  }
+};
+
```
akinfelami, line 76: Move this function into the component it is "serving". Better code structure

akinfelami: Additionally, wrap it in a Tanstack router mutation - so that you can automatically invalidate the cache and the relevant information is re-populated without the need for refreshing the page! It's pretty neat!

akinfelami (changes requested):
Folks, great job on this one. But like you shared, it is incomplete. 

I saved you some trouble. The PR looked like this initially
Screenshot 2024-03-22 at 12 07 35

but now looks like this:

Screenshot 2024-03-22 at 12 28 24

BUT there are still a few things that need tweaking. Kindly take a look at the comments and reach out with any questions

jasozh (approved):

PR #202, opened 2024-03-24
Export database fullstack

hehub000:
## Summary

Added page "website" with a button that allows user to download a csv file with all tables in the database

## Testing

Manually tested all new routes individually and all helper functions individually.

## Notes

The Permissions table is currently empty, so it's omitted from the array of tables to be queried and converted to csv (line 62 of ManageWebsite). It should eventually be added to that array.


frontend/package.json
```diff
@@ -36,6 +36,7 @@
     "react-firebase-hooks": "^5.1.1",
     "react-hook-form": "^7.50.1",
     "react-html-parser": "^2.0.2",
+    "react-json-csv-convert": "^1.1.1",
```
akinfelami, line 39: hey guys - i'm not sure about this one. It has only 28 weekly downloads. Compared to a package like `papaparse` which has about 2M weekly downloads. Sounds like we might have to opt for `papaparse

Please check it out here - https://www.npmjs.com/package/papaparse

frontend/src/components/organisms/ManageWebsite.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const userData = await api.get("/users/")
+    const users = userData.data.data.result
+
+    const preferencesData = await api.get("/users/preferences") // new endpoint
+    const preferences = preferencesData.data.data
+
+    // Combine all data arrays
+    const allData = [events, enrollments, users, preferences];
+    const allCSVs = []
```
akinfelami, line 63: Okay this is going to be a significant change to the overall design of this PR, but i think it is more ideal to keep the logic server side. What I mean is that instead of making different requests (multiple endpoints) you can simply use one endpoint (maybe /download or something) then in the server, you don't even need to make new endpoints - simply use Prisma and it's APIs to make the relevant queries. Then send the combined JSON back as a response

Please reach out if I am not clear enough

akinfelami: I don't imagine there will be a size limit but we can see what the response looks like

backend/src/events/controllers.ts
akinfelami, line 1: take a look at my comment in Comment on lines +43 to +63 in `ManageWebsite`. Refactoring to that effect will render most of these new endpoints and controllers obsolete

frontend/src/components/organisms/ManageWebsite.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const anchor = document.createElement('a');
+    anchor.href = URL.createObjectURL(blob);
+    anchor.download = 'database.csv';
+    anchor.click();
+  
+    // Clean up
+    URL.revokeObjectURL(anchor.href);
+  }
+
+  const handleDownloadDatabase = async () => {
```
akinfelami, line 41: Please when all is refactored - remember to wrap this in useQuery hook, that way we can better manage, loading, data, success and error states

akinfelami (changes requested):
Hi guys!

Great work on this! There is quite a few that will need to be refactored. Please take a look at let me know if there are any questions

jasozh (approved):

PR #208, opened 2024-03-24
Email2 templates frontend

tmavani23:
## Summary

Creates objects that converted all six html files made with SendGrid into strings to integrate into existing functions within backend/src/events and backend/src/users. 
[EDIT 3/28] Made changes to the html strings and paths so that it works now. Also added sendEmail functionality in backend/src/events/controllers.ts for confirmUser to send an email when a user's attendance to an event is confirmed by a supervisor and for addAttendee when a user registers for an event. Also made changes in backend/src/users/controllers.ts to add sendEmail functionality to editStatus and editRole. Email template for blacklisting a user is created and in the emails folder but has not been implemented in a function because that function has not yet been created. 

## Testing

Utilized new version of About.tsx given by Jason to test the emails with test button on localhost.


backend/src/events/controllers.ts
```diff
@@ -353,12 +366,12 @@ const deleteAttendee = async (
...
   cancelationMessage: string
 ) => {
-  // grabs the user and their email for SendGrid fucntionality
+  // grabs the user and their email for SendGrid functionality
   const user = await userController.getUserByID(userID);
   var userEmail = user?.email as string;
 
   // sets the email message
-  const emailHtml = "USER REMOVED FROM THIS EVENT";
+  const emailHtml = "<b>htmlRegCancel</b>";
```
akinfelami, line 374: I don't think this will work. This is simply a string not an html file. Did you test this?

akinfelami (changes requested):
Hey guys!

Great work so far!

But I believe this PR is still incomplete. The different html templates are supposed to server different functions, hence they should be associated with the various controllers that match their functionality. Please address this and reach out with any questions

jasozh (approved):

PR #210, opened 2024-03-24
Map Implementation Done, Errors Due to API Key Being Inactive

brandon-lerit:
## Summary

Map Widget Added to the page. Code written to get the coordinates for the center and marker but not sure if the code is wrong or if the API Key not being active is causing issues.

### DIFF
--- a/frontend/src/components/organisms/ViewEventDetails.tsx
+++ b/frontend/src/components/organisms/ViewEventDetails.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import EventTemplate from "../templates/EventTemplate";
 import EventCardRegister from "./EventCardRegister";
 import Divider from "@mui/material/Divider";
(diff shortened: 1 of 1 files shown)


frontend/src/components/organisms/ViewEventDetails.tsx
```diff
@@ -77,6 +79,44 @@ const ViewEventDetails = () => {
...
+  const { isLoaded } = useJsApiLoader({
+    id: "google-map-script",
+    googleMapsApiKey: process.env.NEXT_PUBLIC_GOOGLE_API_KEY!,
+  });
+
+  const containerStyle = {
+    width: "100%",
+    height: "400px",
+    margin: "auto",
+  };
```
akinfelami, line 95: Please use tailwind inline styles instead

frontend/src/components/organisms/ViewEventDetails.tsx
```diff
@@ -77,6 +79,44 @@ const ViewEventDetails = () => {
 
   const dateHeader = formatDateTimeToUI(datetime);
 
+  setKey(process.env.NEXT_PUBLIC_GOOGLE_API_KEY!);
+  setLanguage("en");
+  setRegion("es");
```
akinfelami, line 84: why is the region set to "es"?

akinfelami (changes requested):
hey guys! Amazing work on this so far. Please address the following:

- include the `package.json` file so that we can install the new packages
- push fixes to the comments!

PR #216, opened 2024-04-05
Refactor AuthContext and other pages

jasozh:
## Summary

Started out as an AuthContext fix and then I started adding random stuff.
- Highlight: Navigating between pages no longer causes a brief white flash that is impossible to prevent. The reason is because before the page is fully loaded, AuthContext flashes a pure white loading screen. This is replaced by moving the loading component to the individual templates, so each template can have their own loading screen that includes the background gradient.
- The fetch data error component is changed slightly so that when it is used, the background is also no longer white but includes the gradient.
- Fix minor styling with the Editor
- Standardize button group order across all modals + forms.
- Add a footer.
- Fix minor bugs in EventCard, ViewEvents, and ViewEventDetails so that the right data displays for supervisor/admin/volunteer. Navbar items also take role into account now.
- Basic password constraints are now enforced on SignupForm, ProfileForm, and ResetPassword Form: 6 chars in length, 1 uppercase, 1 lowercase, 1 special char
- Add placeholder data for Past Events
- Add Edit Event and View Event to Manage Event page

image


frontend/src/utils/AuthContext.tsx
```diff
@@ -207,13 +207,14 @@ export const AuthProvider = ({ children }: AuthProviderProps) => {
...
-    );
-  }
+  // No longer needed, moved to individual templates to avoid white flash on load
+  // if (loading || !isAuthenticated) {
+  //   return (
+  //     <div>
+  //       <Loading />
+  //     </div>
+  //   );
+  // }
```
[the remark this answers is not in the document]
akinfelami: Also, login as a volunteer and try navigate to a protected route like event/create, you will notice the page flashes first (not ideal)

frontend/src/pages/events/[eventid]/cancel.tsx
Comment (akinfelami) [inline frontend/src/pages/events/[eventid]/cancel.tsx:1]: Are you also getting similar erros?

Screenshot 2024-04-06 at 20 21 12

frontend/src/components/templates/CenteredTemplate.tsx
akinfelami, line 1: Out of curiosity, how does moving the hideContent logic to the different templates remove the white flashing thing?

frontend/src/components/organisms/ViewEvents.tsx
```diff
@@ -83,55 +86,42 @@ const UpcomingEvents = () => {
...
-        </div>
-      ))}
-
+      {(role === "Supervisor" || role === "Admin") &&
+        upcomingEventsSupervisor.map((event: ViewEventsEvent) => (
+          <div>
+            <div className="mt-5" />
+            <EventCardNew key={event.id} event={event} />
+          </div>
+        ))}
```
akinfelami, line 116: I don't think the check for the role will be necessary here because the query for the `upcomingEventsSupervisor` is something like this `/events?ownerid=${userid}&date=upcoming&sort=startDate:desc`, notice the ownerid so the array only holds events a user created, which means it will be empty for volunteers?🤔

frontend/src/components/organisms/ViewEvents.tsx
```diff
@@ -83,55 +86,42 @@ const UpcomingEvents = () => {
...
-        </div>
-      ))}
-
+      {role === "Volunteer" &&
+        upcomingEventsVolunteer.map((event: ViewEventsEvent) => (
+          <div>
+            <div className="mt-5" />
+            <EventCardNew key={event.id} event={event} />
+          </div>
+        ))}
```
akinfelami, line 124: Follows from my earlier question on whether this is necessary, if you register for an event, you should be able to View the event details. This applies to admin and supervisors that might want to register for events

frontend/src/components/organisms/ViewEvents.tsx
```diff
@@ -396,22 +405,74 @@ const PastEvents = () => {
...
-        dataSetLength={pastSupervisorEvents.total}
-        paginationModel={paginationModelSupervisor}
-        setPaginationModel={setPaginationModelSupervisor}
-      />
+      {((role === "Admin" && pastSupervisorEvents.result.length == 0) ||
+        (role === "Supervisor" && pastSupervisorEvents.result.length == 0) ||
+        (role === "Volunteer" && pastVolunteerEvents.result.length == 0)) && (
+        <div className="p-10">
+          <div className="text-center">There are no past events</div>
+        </div>
```
akinfelami, line 420: This looks better if it comes after the "reference and certificate hour tracker". Also, it looks like the table is showing anyway if the date is empty, but it is just a white strip that I think isn't that visually appealing

frontend/src/components/organisms/ViewEventDetails.tsx
akinfelami, line 1: not sure if the error for the register page is coming from here (see comments in register page)

frontend/src/components/organisms/SignupForm.tsx
```diff
@@ -186,6 +194,24 @@ const SignupForm = () => {
...
+                hasLower: (value) =>
+                  /.*[a-z].*/.test(value) ||
+                  "Password must contain at least one lowercase letter",
+                hasNumber: (value) =>
+                  /.*[0-9].*/.test(value) ||
+                  "Password must contain at least one number",
+                hasSpecialChar: (value) =>
+                  /.*[\W_].*/.test(value) ||
+                  "Password must contain at least one special character",
+              },
```
akinfelami, line 214: This might not be necessary 🤔. Firebase handles this already, and it is available with our current implementation.

frontend/src/components/organisms/ProfileForm.tsx
```diff
@@ -241,21 +259,22 @@ const ProfileForm = ({ userDetails }: ProfileFormProps) => {
...
               type="button"
               variety="secondary"
               onClick={() => {
-                reset(userDetails, { keepDefaultValues: true });
+                reset(undefined, { keepDefaultValues: true });
               }}
+              disabled={!isDirty}
             >
-              Cancel
+              Reset changes
```
akinfelami, line 275: Could you explain what this is?

frontend/src/components/organisms/ManageAttendees.tsx
akinfelami, line 1: The buttons are a great addition and look great on desktop but I think they might be taking too much "real estate" and in the way on mobile. What do you think about having them as icons in a row on mobile? that way users can quickly access what they came to the manage attendees page to do? 

Screenshot 2024-04-06 at 20 41 57

akinfelami (changes requested):
Hey! Lots of work on this, so many moving parts! Great work!

Added some comments that we might need to discuss further

frontend/src/components/organisms/ViewEvents.tsx
akinfelami: Screenshot 2024-04-06 at 20 53 00

frontend/src/pages/events/[eventid]/register.tsx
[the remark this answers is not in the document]
Comment (jasozh) [reply frontend/src/pages/events/[eventid]/register.tsx:1]: All I did in this file was comment out legacy code. The `register` page does a query fetch, but the results aren't actually used because `ViewEventDetails` does its own query fetch. The error here I noticed as well but it's unrelated to the change I made, although we definitely need to debug this later down the line.

frontend/src/utils/AuthContext.tsx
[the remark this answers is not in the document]
jasozh: So this is actually a surprisingly difficult issue. I spent a few hours researching possible alternatives, and there actually isn't a good solution. Even [professional websites](https://app.splitbee.io/projects/theodorusclarence.com) have the same exact issue we're facing.

The main problem with our current implementation is that having a white flash before every route navigation is very bad from a UI perspective. I tested on both my laptop and phone, and our current website has a half-second white flash over the entire page every time the user navigates to a new route, no exceptions. It makes our website feel very glitchy and unprofessional.

The reason this is happening is because route navigation is instant, but authentication always takes a split second before returning a result. The proper solution to this issue is to implement server-side authentication using the Next.js `middleware.ts` file (see [docs](https://nextjs.org/docs/app/building-your-application/routing/middleware)) so that we intercept route navigation and only return a route once authentication completes. However, I looked into it and implementing this change would require us to refactor too much, because currently Firebase authenticates on the client-side whereas `middleware.ts` requires server-side authentication. We would need to use `firebase-admin` and send our token to the server via cookies using the `nookies` package, etc.

The compromise I decided on was that we would leave authentication handled on the client side, but to make the loading flash appear as invisible as possible. So rather than always loading a white background before every route, the page skeleton loads first and the children load only if the user is actually authenticated. The flash is still there, but because our loading screen looks almost the same as our final result (minus the children), it's effectively invisible.

> Also, login as a volunteer and try navigate to a protected route like event/create, you will notice the page flashes first (not ideal)

Only the page layout flashes. None of the children are actually shown until authentication completes, so no sensitive information is actually shown.

It's not a perfect solution, but it minimizes the tradeoffs we need to make. With the current white loading screen, in-site navigation is very distracting / glitchy (due to the white flash). With this change, in-site navigation is seamless, but the downside is that sometimes the loading screen will seem "wrong' if you directly type in the protected route from the URL. However, users will be navigating with our in-site navigation (buttons + navbar) much more than they will be from the URL bar, so I think it's worth it.

The only proper solution without tradeoffs is to dismantle AuthContext and implement a proper middleware + cookies.

frontend/src/pages/events/[eventid]/cancel.tsx
Comment (jasozh) [reply frontend/src/pages/events/[eventid]/cancel.tsx:1]: The `cancel` page is actually obsolete and needs to be deleted soon. Everything is handled on the `register` page now with `ViewEventDetails`

frontend/src/components/templates/CenteredTemplate.tsx
jasozh: The white flash is because the loading screen in `AuthContext` has a white background while the website waits for authentication to complete. Moving it to different templates means that each template has their own "loading screen" so to speak, which allows us to use the expected color background gradient and so on (instead of just white).

frontend/src/components/organisms/ViewEvents.tsx
jasozh: That makes sense but the reason I actually implemented this was because I was logged in as an admin and I saw two event cards for the same event (one for volunteer and one for the supervisor/admin). I think it's better to be explicit to make sure it doesn't show up.

frontend/src/components/organisms/ViewEvents.tsx
jasozh: I added the conditional here because I think from partner calls they mentioned that supervisors don't register for events. Also the event details are still visible, this conditional is strictly for the registration card in the bottom right.

frontend/src/components/organisms/ViewEvents.tsx
jasozh: Good catch, fixed it.

image

frontend/src/components/organisms/SignupForm.tsx
jasozh: That's interesting because I was totally able to create accounts with password `12345678`. Maybe we just didn't enable it in Firebase yet for testing. Either way, I think there's some value in having `react-hook-form`'s immediate feedback about valid passwords without having to spam signup attempts every single time. This implementation means that the user can correct their password until it's correct without ever having to ping Firebase.

frontend/src/components/organisms/ManageAttendees.tsx
jasozh: That's a good point. I was just going off of the hi-fis in the Figma:

image

I think we should keep it even if it takes up more vertical real estate because the three buttons don't fit on all phone sizes. For example, this is with the iPhone SE screen width:

image

frontend/src/pages/events/[eventid]/register.tsx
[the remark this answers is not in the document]
Comment (akinfelami) [reply frontend/src/pages/events/[eventid]/register.tsx:1]: Sounds good. I will take another look then with that in mind

frontend/src/components/organisms/ManageAttendees.tsx
akinfelami: Would the icons alone (without the words) be unintuitive?

frontend/src/components/organisms/SignupForm.tsx
akinfelami: You're right, however, I think a password of 12345678 is "correct". It is just a courtesy from us for users to have strong passwords. Fwiw, I couldn't find a list of what it means for a weak password, but there is an error code for [weak password](https://firebase.google.com/docs/auth/flutter/password-auth#create_a_password-based_account), I think a password of under 8 characters is what has triggered it for me in the past!

frontend/src/components/organisms/ViewEvents.tsx
akinfelami: Ah right, I wasn't able to see this because of the error I kept seeing on the view events page.

frontend/src/components/organisms/ViewEvents.tsx
akinfelami: Wait, I'm curious about this. It could be because a supervisor created an event, and also registered for it? Which speaks to one of the BUGS mentioned in this PR - [here](https://github.com/cornellh4i/lagos-volunteers/issues/222)

frontend/src/utils/AuthContext.tsx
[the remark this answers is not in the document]
akinfelami: Yep, that is all valid. I thought about this a lot too when we started implementing authentication. I pushed a commit to hide the navbar. Keeping the right of the Welcome template is a nice touch for navigation between authentication paths like /login /signup but since it shows the image first, in a case where someone navigates to one of the those paths when already logged in, the image shows up first before showing the right content. A little bit uneasy about that but this PR Should not be blocked any further because of that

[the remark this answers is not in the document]
akinfelami: Edit: I removed the change to hide the navbar

frontend/src/pages/events/[eventid]/register.tsx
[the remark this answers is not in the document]
Comment (akinfelami) [reply frontend/src/pages/events/[eventid]/register.tsx:1]: Found the issue: has to do with authentication lol. I think this is how it goes:
- [here](https://github.com/cornellh4i/lagos-volunteers/blob/af3af41cbfa9f397c18e2679bbc9237b3c96d4f5/frontend/src/components/organisms/ViewEventDetails.tsx#L37) if you log the `user?.email` you will notice it is always undefined. I think the user object never resolves before the the request is sent to the backend. This is probably because EventTemplate is missing the hideContent logic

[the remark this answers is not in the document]
Comment (akinfelami) [reply frontend/src/pages/events/[eventid]/register.tsx:1]: But if you uncomment the loading logic from the 'AuthContext` you will notice everything returns back to normal

frontend/src/pages/events/[eventid]/cancel.tsx
Comment (akinfelami) [reply frontend/src/pages/events/[eventid]/cancel.tsx:1]: Yeah right totally forgot about that

frontend/src/pages/events/[eventid]/register.tsx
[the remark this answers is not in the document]
Comment (jasozh) [reply frontend/src/pages/events/[eventid]/register.tsx:1]: Good catch. Fixed the error by wrapping `ViewEventDetails` in a `DefaultTemplate` so that the children are not shown until user authentication completes.

frontend/src/utils/AuthContext.tsx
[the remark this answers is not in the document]
jasozh: Yep, this is the same issue as someone who is not logged in will still see the background gradient if they navigate to a protected page. Unfortunately, there is no solution unless we want to implement a proper middleware.

frontend/src/components/organisms/ViewEvents.tsx
jasozh: I'm not really sure, it might be a possibility. I still think it's better to add a safeguard since it doesn't otherwise impact the functionality in any way.

frontend/src/components/organisms/ViewEventDetails.tsx
jasozh: Yep, the issue was that we were trying to grab the `user` object before auth completed. Fixed it (see above)

frontend/src/components/organisms/SignupForm.tsx
jasozh: I think it's good practice to enforce stricter password requirements. These are the same requirements that Google requires (see below).

image

frontend/src/components/organisms/ManageAttendees.tsx
jasozh: Not necessarily but the icons look pretty and I think it's not worth it. What if we wanted to add a fourth button one day or just rename one of the buttons to be more descriptive? It would then have to become a column again.

akinfelami:
Hey, do you know why it looks like this on mobile?
Screenshot 2024-04-07 at 15 45 49

jasozh:
I can't replicate the issue, this is what the page looks like for me:

image

akinfelami (approved):

PR #224, opened 2024-04-06
Adjust Scheduling so that EndDate is the the same as StartDate

akinfelami:
## Summary

As we discussed, this removes endDate altogether, and simplifies things A LOT.

## Testing
Note: Screenshots for reference: you will notice the new layout and that the errors show up under the inputs, not as notifications
Screenshot 2024-04-06 at 19 32 34

Additionally, when you create a new event, you are immediately redirected back to the home page, but you see this: I think it's better implementation
Screenshot 2024-04-06 at 19 33 30

Finally addressed some minor bugs:
- The duplicate event is also adjusted to not include endDate
- The view events is updated so that events you create show the "Manage event" button
- also fixed minor bug where all the events where using the same photo
## Notes
- There is still the bug where the startTime can be in the past.

### DIFF
--- a/frontend/src/components/organisms/EventCardNew.tsx
+++ b/frontend/src/components/organisms/EventCardNew.tsx
@@ -20,11 +20,11 @@ const EventCardContent = ({ event }: EventCardNewProps) => {
   const date = new Date(event.startDate);
   const dateInfo = displayDateInfo(date);
   const url =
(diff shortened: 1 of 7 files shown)


jasozh (approved):
Awesome work! Just fixed some minor padding issues but otherwise looks good, also nice work on getting notifications to show on `/events/view` after creating an event.

PR #226, opened 2024-04-07
Address some bugs in creating events

akinfelami:
## Summary

Bug fixes include:
- can now enter an address if a Google maps location autocomplete does not return anything
- can now enter and edit descriptions - before it was hard coded as hello world even after creating
- correct radio is checked (virtual or in-person) and in-person location shows when editing the event
- fixed overflow issues- before when text was long it would push the right component too far and messes up the styling
- 

## Testing

try to create/ edit an event!

## Notes

Any other bugs I missed? I think it's best to address them all now

### DIFF
--- a/frontend/src/components/atoms/LocationPicker.tsx
+++ b/frontend/src/components/atoms/LocationPicker.tsx
@@ -190,12 +190,14 @@ const LocationPicker = ({
                 </Grid>
                 <Grid
                   item
-                  sx={{ width: "calc(100% - 44px)", wordWrap: "break-word" }}>
+                  sx={{ width: "calc(100% - 44px)", wordWrap: "break-word" }}
+                >
                   {parts.map((part, index) => (
                     <Box
                       key={index}
                       component="span"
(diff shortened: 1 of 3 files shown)


jasozh (approved):
Seems good to me; awesome that you got LocationPicker to work with `react-hook-form`  and `Controller` since I struggled with that over winter break. however since we don't have an API key I can't guarantee that the changes haven't messed up the google maps API, but I'm approving this anyways since we might not even be using google maps api

jasozh:
> Any other bugs I missed? I think it's best to address them all now

There's still some stuff in the backlog although some of the ticket items are a bit outdated now.

akinfelami:
> Seems good to me; awesome that you got LocationPicker to work with `react-hook-form` and `Controller` since I struggled with that over winter break. however since we don't have an API key I can't guarantee that the changes haven't messed up the google maps API, but I'm approving this anyways since we might not even be using google maps api

Sounds good to me!

PR #232, opened 2024-04-14
Automatic data refresh

hehub000:
## Summary

When one instance of the frontend makes a change to the attendee status of an event, it will automatically update for other instances of the frontend.

## Testing

Manually tested using two private browsers on the same computer

## Notes

I noticed that the pattern in backend/src/events/views is that each socketnotify resource is simply the event that is being updated, so when I added a socketnotify to the route relevant to this ticket, I followed the same convention. I'm wondering if we can/should make the socketnotify resource more specific for each route, so that data refreshes are only happening exactly when they need to.

### DIFF
--- a/backend/api-spec.json
+++ b/backend/api-spec.json
@@ -124,6 +124,11 @@
             "type": "string"
           },
           {
+            "name": "attendeeStatus",
+            "in": "query",
+            "type": "string"
+          },
+          {
             "name": "sort",
             "in": "query",
             "type": "string"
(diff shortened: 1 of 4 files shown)


akinfelami (changes requested):
Hi guys! Great work on this. I tested it and it worked!

A few comments. and I agree that the current pattern is a little unintuitive. Let's discuss

akinfelami (approved):

jasozh (approved):

PR #239, opened 2024-04-18
Add Blacklisted tab and count hours accurately

jasozh:
## Summary

- Add Blacklisted tab
- Add accurate hours calculations in
- Edit manage attendees message
- Fix select background
- Fix hours in ManageUsers table

### DIFF
--- a/backend/src/users/controllers.ts
+++ b/backend/src/users/controllers.ts
@@ -118,6 +118,9 @@ const getUsers = async (
   pagination: {
     after: string;
     limit: string;
+  },
+  include: {
+    hours: boolean;
   }
 ) => {
   /* SORTING */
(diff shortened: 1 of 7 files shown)


backend/src/users/controllers.ts
```diff
@@ -240,11 +241,23 @@ const getUsers = async (
...
   });
+
+  // Get hours for each user
+  const newQueryResult: any = queryResult;
+  for (let i = 0; i < queryResult.length; i++) {
+    const user = queryResult[i];
+    const hours = await getHours(user.id);
+    newQueryResult[i].totalHours = hours;
+  }
+
```
akinfelami, line 252: I thought we only want to grab the hours if the query specifies?

frontend/src/components/organisms/ManageUserProfileNew.tsx
```diff
@@ -143,7 +153,7 @@ const ManageUserProfileNew = () => {
     email,
     joinDate,
     userId,
-    hours,
+    hours: fakeHours,
```
akinfelami, line 156: What's this about?

akinfelami (changes requested):
Great!

Two quick comments and we should be able to merge!

frontend/src/components/organisms/ManageUserProfileNew.tsx
jasozh: I can't delete the `hours` column from the `UserProfile` column, so we still grab the hours from the profile but we can't use it. I use `hours` for the actual hours and `fakeHours` for the hours grabbed from profile.

backend/src/users/controllers.ts
jasozh: fixed

akinfelami (approved):
lgtm

PR #242, opened 2024-04-21
Search name fullstack

DiegolMarques:
## Summary

*NOT COMPLETE*
Users can now be searched for by first name, last name, or email, but not full name. (more in notes)

## Testing

Manually tested.

## Notes

It seems like the OR key in the prisma filter dictionary doesn't work the way we expect, but if it did, this ticket would be complete. We can't figure out the specifics of how it works from the documentation; it seems like it's not behaving the same way it does in the documentation. We may be able to change the backend schema to make the objects more suited for this kind of lookup, but we didn't want to make significant changes like those unless only necessary.

### DIFF
--- a/frontend/src/components/organisms/ManageAttendees.tsx
+++ b/frontend/src/components/organisms/ManageAttendees.tsx
@@ -402,6 +402,7 @@ const Pending = () => {
     }
   };
   //first commit
+  //second commit
   /** Tanstack query to fetch attendees data */
   const {
     data: pendingData,
--- a/frontend/src/components/organisms/ManageAttendees.tsx
+++ b/frontend/src/components/organisms/ManageAttendees.tsx
@@ -345,7 +345,16 @@ const ModalBody = ({
   );
 };
 
-const Pending = () => {
+interface TabProps {
+  queryString: string; // Define the type for pendingData
+  statusVar: string;
+  enumVar: "PENDING" | "CHECKED_IN" | "CHECKED_OUT" | "REMOVED" | "CANCELED";
+
+}
+
+
+
+const TabComponent: React.FC<TabProps> = ({ queryString, statusVar, enumVar }) => {
   const router = useRouter();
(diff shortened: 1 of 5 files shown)


frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -562,8 +563,7 @@ const CheckedIn = () => {
     queryKey: ["event", eventid, paginationModel.page, "checked_in"],
     queryFn: async () => {
       const { data } = await api.get(
-        `/users?eventId=${eventid}&attendeeStatus=${"CHECKED_IN"}&limit=${
-          paginationModel.pageSize
+        `/users?eventId=${eventid}&attendeeStatus=${"CHECKED_IN"}&limit=${paginationModel.pageSize
```
akinfelami, line 566: what is this about? I thought this pr was only for searching

akinfelami (changes requested):
Great work folks!

I added a commit that simplified things a bit. But, there is a possibility I introduced some bugs. Please test it accordingly, and let me know if the "incompleteness" remains

akinfelami:
Making this blocked for now. The search functionality works alright, but when you navigate to other pages, and search, things start to break. This is also related to the issue we were facing with the sorting. I have an idea for how to go about it. Will share more

akinfelami (changes requested):
Hey guys!

Good job getting the code down to 600 lines! Reusing components is clever but I think there are still a few thing missing. 

- Open up prisma studio and grab an event that has 60 enrollments. Navigate to manage attendees for that event and you will notice that when you click to the next page of users under pending, things start to break. 
- I want to suggest using better variable names. Maybe remove the `Var` suffixes you have (personal opinion) Also, using queryStrings like `queryString="have removed their registration"` is probably not efficient. It's too "magicy", and more strings = more bytes = more memory. I suggest using one word to describe they keys.

PR #243, opened 2024-04-22
Log in with Google

### LINKED ISSUE #236: Log in with Google (fullstack)
**Branch:** `login-google-fullstack`

### Task at Hand

1. Implement your code in `frontend/src/components/organisms/LoginForm.tsx`.
2. Confirm that your code and pull request follow the [Pull Request Checklist](https://www.notion.so/h4i/Pull-Request-Checklist-910d5e6925df49569067f00f6aa0c0f3).

### Acceptance Criteria

- [x] Users should be able to use the **Continue with Google** button on the login page to sign in with their Google account.
- [x] If the user email does not already exist in the system, doing a Google sign in should create a new user account. There should be no need to send a password verification email since Google is a trusted source, so the user should already be set to verified.
- [x] If the user email does exist in the system, doing a Google sign in should act as a log in. Users should be able to sign in with both their email and password, and the auth providers should be linked.

### Notes

- See the Firebase documentation for Google authentication [here](https://firebase.google.com/docs/auth/web/google-signin).
- See the documentation for linking auth providers [here](https://firebase.google.com/docs/auth/web/account-linking).
- Maybe [this](https://blog.bitsrc.io/firebase-authentication-with-react-for-beginners-implementing-email-password-and-google-sign-in-e62d9094e22) is helpful.

owenchen04:
## Summary

Allowed users to create an account and sign-in using the 'Continue with Google' button. The flow is the same regardless of whether they are using Google to log in for the first time or they are a returning user.

## Testing

Manually tested by using my personal Google account. Made sure that in both cases (first time signing in vs. returning user), continue with Google worked as expected (creating a user in the local database in the former case). Also used console.log to ensure that the created firebase user had emailVerified already set to true.

## Notes

I was having issues using the react-firebase-hooks useSignInWithGoogle hook, as the googleUser object returned by the hook was always undefined. Instead, I followed the official firebase documentation. While the flow seems fine for now, there is a possibility that there are errors I couldn't account for.

Closes #236

### DIFF
--- a/backend/src/users/views.ts
+++ b/backend/src/users/views.ts
@@ -75,6 +75,47 @@ userRouter.post(
   }
 );
 
+userRouter.post(
+  "/googleCreate",
+  NoAuth as RequestHandler,
+  async (req: Request, res: Response) => {
+    // #swagger.tags = ['Users']
+    socketNotify("/users");
+    let user;
+    const { ...rest } = req.body;
+    try {
+      // Create user in local database
+      user = await userController.createUser(
+        rest,
+        rest.profile,
+        rest.preferences,
+        rest.permissions
+      );
+
+      // If local user doesn't exist, throw error
+      if (!user) {
+        throw Error("Failed to create local user");
+      }
+
+      // If test environment, return user without firebase auth
+      else if (process.env.NODE_ENV === "test") {
+        return res.status(201).send({ success: true, data: user });
+      }
+      
+    } catch (e: any) {
+      // If user exists in local database but not in Firebase, delete user
+      if (user) {
+        try {
+          await userController.deleteUser(user.id);
+        } catch (e: any) {
+          return res.status(500).send({ success: false, error: e.message });
+        }
+      }
+      return res.status(500).send({ success: false, error: e.message });
+    }
+  }
+);
+
 userRouter.delete("/:userid", useAuth, async (req: Request, res: Response) => {
   // #swagger.tags = ['Users']
   attempt(res, 200, () => userController.deleteUser(req.params.userid));
--- a/backend/src/users/views.ts
+++ b/backend/src/users/views.ts
@@ -76,26 +76,25 @@ userRouter.post(
 );
 
 userRouter.post(
-  "/googleCreate",
+  "/create",
   NoAuth as RequestHandler,
   async (req: Request, res: Response) => {
     // #swagger.tags = ['Users']
     socketNotify("/users");
     const { ...rest } = req.body;
-    
     try {
       await firebase.auth().setCustomUserClaims(rest.id, {
         admin: false,
         supervisor: false,
         volunteer: true,
       });
-  
-      attempt(res, 200, () => userController.createUser(
+      const user = await userController.createUser(
         rest,
         rest.profile,
         rest.preferences,
         rest.permissions
-      ));
+      );
+      return res.status(200).send({ success: true, user: user });
     } catch (e: any) {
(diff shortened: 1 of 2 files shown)


backend/src/users/views.ts
```diff
@@ -75,6 +75,47 @@ userRouter.post(
...
+          await userController.deleteUser(user.id);
+        } catch (e: any) {
+          return res.status(500).send({ success: false, error: e.message });
+        }
+      }
+      return res.status(500).send({ success: false, error: e.message });
+    }
+  }
+);
+
```
akinfelami, line 118: You don't need all this. Since it's just one action, simply use the patterns of `attempt` that other endpoints use

akinfelami: You will actually need to set the user's custom claims in this controller

frontend/src/components/organisms/LoginForm.tsx
```diff
@@ -109,8 +112,67 @@ const LoginForm = () => {
   }, [signInErrors]);
 
   /** Sign in with Google */
-  const [signInWithGoogle, googleUser, googleLoading, googleError] =
-    useSignInWithGoogle(auth);
+
+  const handleGoogleLogin = async () => {
+    const provider = await new GoogleAuthProvider();
```
akinfelami, line 117: no need to await this?

frontend/src/components/organisms/LoginForm.tsx
```diff
@@ -109,8 +112,67 @@ const LoginForm = () => {
...
+        const { response, data } = await api.get(`/users/${userid}`);
+
+        if (!data["data"]) {
+          // console.log("creating user");
+          await createLocalUserFromGoogle.mutateAsync({userid, firstName, lastName, email, phoneNumber});
+        } else {
+          // console.log("user already created");
+        }
+      }
+      createUser();
```
akinfelami, line 139: createUser is an async function but you don't await this or use then/catch

frontend/src/components/organisms/LoginForm.tsx
```diff
@@ -109,8 +112,67 @@ const LoginForm = () => {
...
+        const [firstName, lastName] = user?.displayName ? user?.displayName?.split(" ") : ["", ""]
+        const email = user?.email as string;
+        const userid = user?.uid as string;
+        const phoneNumber = user?.phoneNumber as string;
+        // Check if user exists in local database
+        const { response, data } = await api.get(`/users/${userid}`);
+
+        if (!data["data"]) {
+          // console.log("creating user");
+          await createLocalUserFromGoogle.mutateAsync({userid, firstName, lastName, email, phoneNumber});
```
akinfelami, line 134: this can potentially throw an erorr but there is no error management for it

akinfelami (changes requested):
Hi guys!

Nice work on this. I tested it and it works fine! However, I noticed some things that will need your attention. Please see my comments

akinfelami (approved):

jasozh:
Thinking about potential issues:

1. User creates email/password account then Google account. No issue here, the Google account automatically links to the email/password account in Firebase. ISSUE: "Your account already exists. Please log in using your email and password." error never triggers. Consider removing? Having an automatic link is desired behavior imo.
2. User creates Google account thatn email/password account. Resetting their password is the only way to change their password (to create an email/password account), and doing so also replaces their Google account in Firebase with email/password. But no user-facing issue, because signing into Google automatically links their accounts again. So maybe a non-issue?

jasozh (approved):

PR #244, opened 2024-04-22
Email notifications setting using checkbox

tmavani23:
## Summary

Added a new function in ProfileForm.tsx called updatePreferencesInDB which takes the changed data from the checkbox and edits that users preferences. We then added if statements in all the sendEmail function calls within events/controllers.ts and users/controllers.ts to check if the user's email notifications are on or off. It only sends the email if the user's notifications are on. 

## Testing

I used my log in and selected that I did not want email notifications and then I registered for an event. I did not receive an email so it appears to be working.

### DIFF
--- a/backend/src/users/controllers.ts
+++ b/backend/src/users/controllers.ts
@@ -492,6 +492,7 @@ const getUserPreferences = async (userId: string) => {
   return prisma.user.findUnique({
     where: {
       id: userId,
+      // sendEmailNotification: emailNotifications,
     },
     include: {
       preferences: true,
@@ -507,7 +508,8 @@ const getUserPreferences = async (userId: string) => {
  */
 const editPreferences = async (
   userId: string,
-  preferences: UserPreferences
+  preferences: UserPreferences,
+  // emailNotifications: boolean
 ) => {
   return prisma.user.update({
     where: { id: userId },
@@ -515,6 +517,7 @@ const editPreferences = async (
       preferences: {
         update: {
           ...preferences,
+          // sendEmailNotification: emailNotifications,
         },
       },
     },
@@ -546,7 +549,10 @@ const editStatus = async (userId: string, status: string) => {
         userName,
         textBody
       );
-      await sendEmail(userEmail, "You have been blacklisted.", updatedHtml);
+      const userPreferences = await userController.getUserPreferences(userId);
+      if (userPreferences?.preferences?.sendEmailNotification == true) {
+        await sendEmail(userEmail, "You have been blacklisted.", updatedHtml);
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


backend/src/users/controllers.ts
```diff
@@ -507,14 +508,16 @@ const getUserPreferences = async (userId: string) => {
  */
 const editPreferences = async (
   userId: string,
-  preferences: UserPreferences
+  preferences: UserPreferences,
+  // emailNotifications: boolean
```
akinfelami, line 512: remove comment

backend/src/users/controllers.ts
```diff
@@ -492,6 +492,7 @@ const getUserPreferences = async (userId: string) => {
   return prisma.user.findUnique({
     where: {
       id: userId,
+      // sendEmailNotification: emailNotifications,
```
akinfelami, line 495: remvoe comment

backend/src/users/controllers.ts
```diff
@@ -507,14 +508,16 @@ const getUserPreferences = async (userId: string) => {
...
+  preferences: UserPreferences,
+  // emailNotifications: boolean
 ) => {
   return prisma.user.update({
     where: { id: userId },
     data: {
       preferences: {
         update: {
           ...preferences,
+          // sendEmailNotification: emailNotifications,
```
akinfelami, line 520: remove comment

frontend/src/components/organisms/ProfileForm.tsx
```diff
@@ -89,7 +91,7 @@ const ProfileForm = ({ userDetails }: ProfileFormProps) => {
       oldPassword: "",
       newPassword: "",
       confirmNewPassword: "",
-      emailNotifications: false,
+      emailNotifications: true,
```
akinfelami, line 94: why is the default case change to true?

frontend/src/components/organisms/ProfileForm.tsx
```diff
@@ -98,7 +100,8 @@ const ProfileForm = ({ userDetails }: ProfileFormProps) => {
   // TODO: Implement this
   const [checked, setChecked] = useState(false);
   const handleCheckbox = () => {
-    setChecked((checked) => !checked);
+  setChecked ((checked) => !checked);
+  setEmailNotifications((prev) => !prev);
```
akinfelami, line 104: hmm I think you already have the information you need in the `checked` state

akinfelami (changes requested):
Great work guys!

I have some comments for the review:
- Yes, you tested that you did not receive an email after unchecking the box, but can you confirm that it works the other way around ? so when the box is checked you receive an email?
- I'm curious how you tested this because I am unable to without having to change passwords. Did you encounter this? If so, did the emailNotifications state persist on refresh?

akinfelami (approved):

jasozh (commented):
BUG: Checkbox does not show the actual state on page load, always remains unchecked.
BUG: Checkbox needs to be clicked then unclicked. Leaving it empty does not change the state of the email preferences.

===

FIXED: Profile uses `/users` instead of `/users/search`

PR #245, opened 2024-04-22
Added Tanstack Mutation for canceling events and modal

dvala041:
## Summary

1. Created handleCancelEventAsync tanstack mutation in EventForm.tsx for calling event endpoint and cancelling events
2. Created modal for cancelling an event. On click it calls our handleCancelEvent function

## Testing
1. Tested by running website
2. also tested endpoint with postman

## Notes
1. Cancelling event sends admin back to home page (we are trying to fix this)
2. The "cancel" and "save changes" buttons do not stay disabled because the page refreshes


akinfelami (approved):

jasozh:
BUG: Cancel event button is grayed out even if event is upcoming and is not canceled.

jasozh:
Add cancel banner on top
image

jasozh (approved):

PR #246, opened 2024-04-22
Prevent MDX editor from lazy loading

jasozh:
## Summary

MDX Editor currently lazy loads which looks bad since there's a pause before the component appears. From the MDX docs [here](https://mdxeditor.dev/editor/docs/getting-started), dynamic import is not actually necessary in the pages router, so I'm importing it normally instead.


akinfelami (approved):
lgtm

PR #247, opened 2024-04-22
Prevent event creation in the past

jasozh:
## Summary

Fixes a current bug where supervisors can create events that are in the past if they happen on the same day if they select a time that is earlier than now (e.g. if today is 4/13 5:00 pm, supervisors can still create an even that happens on 4/13 from 8:00 am to 9:00 am). Solution: make the minimum event date tomorrow (also means we don't need to do more cumbersome date validation).

### DIFF
--- a/frontend/src/components/atoms/DatePicker.tsx
+++ b/frontend/src/components/atoms/DatePicker.tsx
@@ -20,9 +20,6 @@ const CustomDatePicker = forwardRef(
     { label, error = "", value, ...props }: DatePickerProps,
     ref: Ref<HTMLInputElement>
   ) => {
-    const today = dayjs();
-    const tomorrow = dayjs().add(1, "day");
-
     return (
       <div>
         <div className="mb-1">{label}</div>
--- a/frontend/src/components/atoms/DatePicker.tsx
+++ b/frontend/src/components/atoms/DatePicker.tsx
@@ -20,6 +20,9 @@ const CustomDatePicker = forwardRef(
     { label, error = "", value, ...props }: DatePickerProps,
     ref: Ref<HTMLInputElement>
   ) => {
+    const today = dayjs();
+    const tomorrow = dayjs().add(1, "day");
+
     return (
       <div>
         <div className="mb-1">{label}</div>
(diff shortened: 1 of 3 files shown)


akinfelami:
Doesn't this mean that they wouldn't be able to create events on the same day? 
There is an alternative prop that we can pass into the timepicker that will prevent times before the current time. See [here](https://mui.com/x/react-date-pickers/validation/#past-and-future-validation:~:text=The%20disablePast%20prop%20prevents%20the%20selection%20all%20values%20before%20today%20for%20date%20pickers%20and%20the%20selection%20of%20all%20values%20before%20the%20current%20time%20for%20time%20pickers.%20For%20date%20time%20pickers%2C%20it%20will%20combine%20both.)

jasozh:
That means we would need to scrap the current implementations of `DatePicker` and `TimeField` and instead use a unified `DateTimePicker` and grab the date + time from that picker. That's more work and we can definitely consider it down the line, but as of right now I think the fix is needed because the current implementation of creating events in the past is just broken.

akinfelami:
> Doesn't this mean that they wouldn't be able to create events on the same day? There is an alternative prop that we can pass into the timepicker that will prevent times before the current time. See [here](https://mui.com/x/react-date-pickers/validation/#past-and-future-validation:~:text=The%20disablePast%20prop%20prevents%20the%20selection%20all%20values%20before%20today%20for%20date%20pickers%20and%20the%20selection%20of%20all%20values%20before%20the%20current%20time%20for%20time%20pickers.%20For%20date%20time%20pickers%2C%20it%20will%20combine%20both.)

Actually just noticed that disablePast was the prop that was removed

akinfelami:
I pushed a commit that:
- changes TimeField to TimePicker (surprisingly, it was a drop-in replacement)
- use `disablePast` for the TimePicker component too

akinfelami (approved):
Porting to Timepicker didn't seem to break anything

jasozh:
Finally implemented proper solution which is just to create a new `timeAndDateValidation()` function that checks to see (before submission) if the input date is before the current datetime.

PR #248, opened 2024-04-22
Prevent past events from being edited

jasozh:
## Summary

When a supervisor tries to edit an event that happened in the past, the Save changes button and Cancel event button are disabled, and an alert is shown instead.

## Testing

image

### DIFF
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -183,6 +183,14 @@ const getEvents = async (
  * @returns promise with eventID or error.
  */
 const updateEvent = async (eventID: string, event: Event) => {
+  const eventDetails = await getEvent(eventID);
+  const currentDate = new Date();
+  if (eventDetails) {
+    if (eventDetails.startDate < currentDate) {
+      return Promise.reject("Event has already started");
+    }
+  }
+
   return prisma.event.update({
     where: {
       id: eventID,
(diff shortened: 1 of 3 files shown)


akinfelami:
Using string type instead of date type will break things in different places because the helpers function to converttoIso is being used by different components.

akinfelami (commented):
lgtm. I changed the types back to Date because it was breaking deployment. I tested it and it still worked. Can you share why you reverted to string?

jasozh:
The reason I changed it to `string` is because the current date type is actually lying; it's not actually a `Date` object but a string. If the parameters to `convertToISo` were actually `Date` objects, there would be no need to do `new Date(inputDate)` because `inputDate` would already be a `Date`. TypeScript doesn't actually enforce types on runtime so it just passes.

akinfelami:
A

> The reason I changed it to `string` is because the current date type is actually lying; it's not actually a `Date` object but a string. If the parameters to `convertToISo` were actually `Date` objects, there would be no need to do `new Date(inputDate)` because `inputDate` would already be a `Date`. TypeScript doesn't actually enforce types on runtime so it just passes.

I see. I did notice it was breaking stuff in other components that actually passed in `Date` objects to the `convertToIso` function. I am fine with either but it will take some effort to refactor others

akinfelami (approved):

akinfelami:
btw I added backend protection in this pr

PR #254, opened 2024-04-26
Profile account settings (redesign-profile-fullstack)

tunile943:
## Summary

Decouple the profile form so now users can submit profile changes and password changes independently from one another.

image
image
image

## Testing

We manually test it out on the front end, then check to see if back end is updated properly.

## Notes

We did comment out validation for strong password during testing, so that might be something that we might want to add again in the future.

### DIFF
--- a/frontend/src/components/organisms/ChangePasswordForm.tsx
+++ b/frontend/src/components/organisms/ChangePasswordForm.tsx
@@ -1,35 +1,24 @@
 import React, { useState } from "react";
 import Button from "../atoms/Button";
 import TextField from "../atoms/TextField";
-import Checkbox from "../atoms/Checkbox";
 import { auth } from "@/utils/firebase";
 import { useForm, SubmitHandler } from "react-hook-form";
 import { reauthenticateWithCredential, EmailAuthProvider } from "firebase/auth";
 import Snackbar from "../atoms/Snackbar";
-import { api } from "@/utils/api";
-import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { useMutation } from "@tanstack/react-query";
 import { updatePassword } from "firebase/auth";
 import { User } from "firebase/auth";
 
 type FormValues = {
   email: string;
-  firstName: string;
-  lastName: string;
-  phoneNumber: string;
-  // preferredName: string;
   oldPassword: string;
   newPassword: string;
   confirmNewPassword: string;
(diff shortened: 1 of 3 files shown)


frontend/src/components/organisms/ChangePasswordForm.tsx
```diff
@@ -0,0 +1,208 @@
...
+            //     /.*[a-z].*/.test(value) ||
+            //     "Password must contain at least one lowercase letter",
+            //   hasNumber: (value) =>
+            //     /.*[0-9].*/.test(value) ||
+            //     "Password must contain at least one number",
+            //   hasSpecialChar: (value) =>
+            //     /.*[\W_].*/.test(value) ||
+            //     "Password must contain at least one special character",
+            // },
+          })}
```
akinfelami, line 172: Why leave this as comments?

frontend/src/components/organisms/ProfileForm.tsx
```diff
@@ -58,16 +53,6 @@ const ProfileForm = ({ userDetails }: ProfileFormProps) => {
         return "User with this email has been disabled.";
       case "user-not-found":
         return "There is no user with this email address.";
```
akinfelami, line 55: Are any of these errors still needed? I don't think so

akinfelami: The argument of `setErrorMessage` after updating the profile should change to the default "something went wrong, try again"

akinfelami (changes requested):
Hi guys!

Great work on this one!

Please see my comments. Additionally, please add the loading state prop to the buttons as usual.

frontend/src/components/organisms/ProfileForm.tsx
tunile943: Just to confirm, I should remove user-disabled and user-not-found right? And for the default message, do you want us to change from "Something went wrong. Please try again" to "Something went wrong, try again."?

akinfelami: The errors listed here are errors gotten from firebase, since you have moved firebase password reset to another component, I don't imagine they will be useful here. However, you  have to be sure removing it altogether does not break anything else. e.g we still need error and success states for changing the user's profile in the db

jasozh (approved):

PR #257, opened 2024-05-05
Fix edit page refresh

ArushiAggarwal:
## Summary

We added UpdateEventData that had the latest information in order to populate the data with since it was returning response in which structure of response might not match what React Query expects. We then used UpdateEventData in  the onSuccess function and switched it from invalidateQueries to setQueryData so we update the corresponding query in the React Query cache. 

## Testing

We created a fake event calling Testing event bug that allowed us to experiment and test our code with the edit event function. We also created console.log comments to for updatedEventData to see if the right data was being passed through.

### DIFF
--- a/frontend/src/components/organisms/EventForm.tsx
+++ b/frontend/src/components/organisms/EventForm.tsx
@@ -202,7 +202,20 @@ const EventForm = ({
           capacity: +volunteerSignUpCap,
           mode: `${mode}`,
         });
-        return response;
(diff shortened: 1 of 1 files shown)


akinfelami (changes requested):
Hey guys, I see you have made some changes to this issue but it doesn't solve the bug. 

If you navigate to the edit page, say from the manage attendees page, the problem persists. 

kindly take another look.

PR #260, opened 2024-06-15
Refactor & Fix all pagination, sorting and searching on tables

akinfelami:
## Summary

Okay this is pretty big. But I refactored ALL the components that used a table to use the new logic for pagination. This implementation works. Pagination, sorting, searching on ANY table. It is also preserves history between components. I believe it addresses all the edge cases we were worried about but please test to crosscheck!

This is also huge because it closes
- https://github.com/cornellh4i/lagos-volunteers/pull/192
- https://github.com/cornellh4i/lagos-volunteers/pull/242

## Testing
Check every component that uses a table and click through. I suggest:
- Move back and forth between pages
- Move to a page deep into pagination and try sort. Tables can sort by name and email. The expected behavior is:
    - pagination resets to page 
    - Pagination restarts with sorted data (still server side)
- try search
   - Searching is improved. Try first name only, last name only (even just some letters in them), try both and try emails
 
Just some ideas on ways to test.

## Notes
- The searching works but can be improved. Instead of having the user have to press enter when they want to search and have to clear the search box when they are done, we can simply make requests based on what they have searched so far but maybe every few seconds so we don't spam the server? What do you think?
- I also changed the API quite a bit. Feel free to revise
- 👀 I can't confirm this doesn't introduce any regression


jasozh:
Bugs fixed:
- After merging, the wrong info was showing on ViewEvents. Turns out in a previous ticket I changed `fetchIdFromDatabase` to point to `/users` instead of `/users/search`, except the pure email filter was never actually implemented. So it was grabbing from the wrong user. Fixed.
- It seems like the edit page refresh bug was fixed in this ticket. I spent way too long on this issue so I might as well write my findings: first I found a workaround and made a PR for that, then found out that the bug was already fixed here, so narrowed it down to a commit from a really long time ago that created a new tanstack query in `ManageAttendees` with the same queryKey as the one used in `edit.tsx`, which was causing the issue. This ticket already fixes it though so closing my previous PR.

jasozh (approved):

jasozh:
NOTE: this is bugged cause `GridSlots` doesn't exist until data grid v7.0 (we only have 6.0 and it seems like there's dependency issues with upgrading)

PR #261, opened 2024-06-15
Allow users to manage authentication providers

jasozh:
## Summary

Users can now see whether they're logged in with Google vs email/password. If logged in with Google, they can create a new password. They can also choose to unlink their Google auth provider (only if they also have email/password enabled). You can no longer "change your password" if you don't have a password set.

## Testing

See screenshots:
image
image

## Notes

User can still reset their password if they don't have a previous password set. This behavior is consistent with other apps (tested with Asana).


akinfelami (commented):
lgtm

PR #264, opened 2024-06-16
Implement volunteer cancel event flow

jasozh:
## Summary

Fixes the following:
- Correctly show that an event is canceled on the ViewEventDetails page.
- BUG: Attempting to cancel registration for a registered event returns an Unauthorized PUT request for volunteers. (likely because of backend API protection preventing volunteers)
- Update how many hours left on cancel event card.
- Volunteers should see their checked in status for upcoming events (see in ViewEventDetails)

## Testing

image

### DIFF
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -463,7 +463,7 @@ const deleteAttendee = async (
       },
     },
     data: {
-      canceled: true,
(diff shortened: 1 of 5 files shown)


akinfelami:
Looks good to me. Pushed a commit to show the user how many days left instead of hours left if it is more than 48 hours

PR #265, opened 2024-06-19
Implement signup cap in ManageAttendees and ViewEventDetails

jasozh:
## Summary

Show the volunteer signup cap properly and prevent volunteers from signing up when exceeding the limit. Also implement WebSockets so that the number of registered volunteers automatically refreshes when the user cancels their registration or is canceled by the supervisor. Finally, volunteers can no longer cancel if the supervisor moves them away from PENDING.

Closes:
- Prevent volunteers from signing up if the volunteer signup cap is already reached

## Testing

image
image

### DIFF
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -344,7 +344,7 @@ const getAttendee = async (eventID: string, userID: string) => {
  * volunteers are registered for the event. Note: volunteers are not considered
  * registered if they have canceled their registration or if their registration
  * has been removed by the supervisor.
- * @param attendees - The list of EventEnrollment objects
+ * @param eventid - The event id
  * @returns The number of registered volunteers
  */
 export const getRegisteredVolunteerNumberInEvent = async (eventid: string) => {
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -279,6 +279,7 @@ const getEvent = async (eventID: string) => {
       id: eventID,
     },
     include: {
+      attendees: true,
       owner: {
         select: {
           profile: true,
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -279,7 +279,6 @@ const getEvent = async (eventID: string) => {
       id: eventID,
     },
     include: {
-      attendees: true,
       owner: {
         select: {
           profile: true,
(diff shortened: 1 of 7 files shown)


backend/src/events/controllers.ts
```diff
@@ -279,6 +279,7 @@ const getEvent = async (eventID: string) => {
       id: eventID,
     },
     include: {
+      attendees: true,
```
akinfelami, line 282: the comment to you pr also applies here

frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -115,6 +119,10 @@ const AttendeesTable = ({
     },
     retry: false,
     onSuccess: () => {
+      // Invalidate the Event Recap query to fetch new data
+      queryClient.invalidateQueries({ queryKey: ["event"] });
```
akinfelami, line 123: Was this key created anywhere?

frontend/src/components/organisms/EventCardCancel.tsx
```diff
@@ -136,32 +139,40 @@ const EventCardCancel = ({
...
-            )}
-          </IconText>
-        )}
         <div className="mt-3" />
         {hoursLeftToCancel < 0 ? (
           <div>
-            You have passed the window for canceling your registration.
-            Registration must be canceled at least 24 hours before the event
-            begins. Failure to do so may affect your volunteer status.
+            have passed the window for canceling your registration. Registration
```
akinfelami, line 152: nit- missing 'You'

akinfelami:
This is good. The websockets are also clutch. However, I think we need to add server side to complete it.

akinfelami (changes requested):

jasozh:
Moved client-side filtering to the backend and created 2 new API endpoints. This means we can get rid of calling the entire attendees object.

jasozh:
> This is good. The websockets are also clutch. However, I think we need to add server side to complete it.

What do we need to implement on the server-side? Tested it in two windows and it seems to work as is.

akinfelami:
> > This is good. The websockets are also clutch. However, I think we need to add server side to complete it.
> 
> What do we need to implement on the server-side? Tested it in two windows and it seems to work as is.

Basically, validations like you can't sign up because the event is full. The server should throw an error if that's the case

jasozh:
Added server-side validation. Merging the PR as is but lemme know if you have any other concerns.

PR #267, opened 2024-06-19
Display cancelation message in ManageAttendees

jasozh:
## Summary

Cancelation messages show up now in a modal according to the hi-fis.

Closes:
- [BUG] Add volunteer cancellation messages to ManageAttendees

## Testing

image

### DIFF
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -279,7 +279,6 @@ const getEvent = async (eventID: string) => {
       id: eventID,
     },
     include: {
-      attendees: true,
       owner: {
         select: {
           profile: true,
--- a/backend/src/events/controllers.ts
+++ b/backend/src/events/controllers.ts
@@ -279,6 +279,7 @@ const getEvent = async (eventID: string) => {
       id: eventID,
     },
     include: {
+      attendees: true,
       owner: {
         select: {
           profile: true,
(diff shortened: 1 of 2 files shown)


backend/src/events/controllers.ts
```diff
@@ -279,6 +279,7 @@ const getEvent = async (eventID: string) => {
       id: eventID,
     },
     include: {
+      attendees: true,
```
akinfelami, line 282: this is tricky. The manage attendees page does a request of the same type (/event/:eventid) to grab the details for a specific event. But because this flag is set - it returns the attendees with it...which is not needed. The attendees list can grow really large. The screenshot below shows 60.

Screenshot 2024-06-19 at 22 17 47

frontend/src/components/organisms/ManageAttendees.tsx
```diff
@@ -72,10 +74,36 @@ type FormValues = {
...
+}
+
+const ViewCancelMessageModalBody = ({
+  handleClose,
+  eventData,
+  userid,
+}: ViewCancelMessageModalBodyProps) => {
+  const eventAttendance = eventData?.attendees?.find(
+    (attendee: any) => attendee.userId === userid
+  );
```
akinfelami, line 90: If this is all we need the eventData for, perhaps we could just make a request that gives us the status of the user from the server?

akinfelami (changes requested):

backend/src/events/controllers.ts
jasozh: I agree. I needed the full attendees object for a couple of reasons but I thought about it and realized I could just make new API endpoints that do the same thing. Fixed it on my other PR and I'll merge the changes over here once that's merged.

frontend/src/components/organisms/ManageAttendees.tsx
jasozh: Yep, made a new endpoint (in the other branch) called `GET /events/:eventid/attendees/:userid`

